PR #630, opened 2015-01-12
Redesign of splashscreen

mrnovvad:
This will be better if someone will change font color of version to white. Sorry for my English


mrnovvad:
Preview
cl-splash

eranif:
Did you check the updated the splash screen? (the one in your branch is different than the current one)
In your current splash, the "CodeLite" title is not set to "bold" and the background text is too visible
Here is the current git head splash for reference (it is under ./Runtime/images/screenshot.png):

splashscreen

RickCHodgin:
It should also include 2015.

mrnovvad:
Fixed all
cl-splash

RickCHodgin:
FWIW, I like Eran's lighter-color version better than the darker version.  For a splash screen it is more similar to the bright and lively colors of the blue/green/yellow splash screen we're used to seeing.  Maybe there could be an association with the theme such that it is an option, or that lighter themes relate to the lighter splash screen, and darker to the darker... ?

mrnovvad:
"Maybe there could be an association with the theme such that it is an option, or that lighter themes relate to the lighter splash screen, and darker to the darker... ?"
Good idea, but this is a work for programmers

RickCHodgin:
I am a programmer. I'll see about adding it (if it's something that's acceptable to Eran).

mrnovvad:
Cool. Glad to see that my first contribution to the OS is don't being left out :)

eranif:
> "Maybe there could be an association with the theme such that it is an option, or that lighter themes 
> relate to the lighter splash screen, and darker to the darker... ?"
> Good idea, but this is a work for programmers

The problem is that the way I have implemented themes is that it is per file type, i.e. you could have the dark "Dracule" theme for C++ workspace,  while using "Back to school" (green theme) for PHP workspace, or just the default and the workspace loading is determined _after_ the splash screen is shown ...

mrnovvad:
May be this is silly, but is it possible to store last workspace theme in a little single file and load it before splash screen?

creatorfromhell:
>Maybe this is silly, but is it possible to store last workspace theme in a little single file and load it before splash screen?
I feel that this wouldn't be an efficient way to go about this. When you start creating new files just to determine what splash screen to use I begin to question if a new splash screen is even wanted or needed.

eranif:
TBH, I like the current splash screen ... and I really think its a matter if taste here.
In order for this PR to be accepted it should be accompanied with a code that allows the user to select the splash screen (which seems a bit overkill for a splash screen...)

PR #697, opened 2015-03-10
Fix building with mingw.

Alexpux:


eranif:
I have applied parts of your pull request (except for gnu++11 and wrong guard for winsock2, I wrote comments on these commits separately)

PR #742, opened 2015-04-28
visual studio solution support

jpcb:
This is support import for visual studio solution version 6,7,8,9,10,11,12 and additional Dev-C++, Borland C++ Builder 6

Concerning to the issues:

Visual Studio 2013 solutions import into empty workspace #460

PR #747, opened 2015-05-02
Import MS Visual Studio solution file (Fixes and improvements)

jpcb:
*Fixes and improvements
*Add CodeBlocksImporter

Reference;
Support-Import Project from Code::Block #746
visual studio solution support #742


eranif:
Thanks, I have merged this and gave it a test on a C::B project file I have lying around here.
It worked (mostly).
One problem I found was the usage of C::B globals (is this the correct terms?) they have something like:
`#wx.lib` since CodeLite uses Makefiles, this breaks the makefile (`#` is a comment)
Maybe a good idea will be to try and locate all the global variables used in the imported solution and prompt the user for replacement? 

Again, thanks for your work so far!

jpcb:
I keep making improvements and corrections, I will be reviewing what you say.

RickCHodgin:
jpcb, thank you for your contributions. :-)

PR #1034, opened 2015-11-23
Tab refill

kluete:
after closing a notebook tab, if there's an empty space in the tab bar that could be re-filled by shifting existing tabs to the right (revealing a former out-of-scope tab) , then do so.

Test this by opening more tabs than can be shown at one time, activate the rightmost tab, then close one of the previous tabs by clicking on it with the middle mouse button. A previously hidden tab should shift back into view.

### DIFF
--- a/Plugin/Notebook.cpp
+++ b/Plugin/Notebook.cpp
@@ -855,8 +855,17 @@ void clTabCtrl::DoUpdateCoordiantes(clTabInfo::Vec_t& tabs)
 
 void clTabCtrl::UpdateVisibleTabs()
 {
+    // force list update if there's a gap that could be refilled
+    const wxRect client_rc(GetClientRect());
+    const wxRect vis_rc = m_visibleTabs.empty() ? wxRect{} : m_visibleTabs.back()->GetRect();
+    
+    const bool  vacancy = (client_rc.GetRight() > vis_rc.GetRight());
+    const bool  refillable = (m_tabs.size() > m_visibleTabs.size());
+    
+    const bool  force_f = vacancy && refillable;
+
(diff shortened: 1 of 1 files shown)


kluete:
(sorry my patch was out-of-date, just rebased to your master)

kluete:
rebased after removing the wxSTC_INDIC_TEXTFORE change.
please check the hardcoded wxRect inflate/deflate hardcoded values
thx!

eranif:
The tabs are now "jumpy". It looks like that this fix also "insist" of displaying a full active tab. i.e. if the I select the last tab (which is partially displayed in the far right) it will become fully visible and by doing this it will shift all tabs to the LEFT. clicking on the tab LEFT to it, will shift all tabs back to the RIGHT

eranif:
Also, I just tested the code without your pull request and it works ... the tab bar is _always_ full
I am going to revert this commit for now.

kluete:
>Also, I just tested the code without your pull request and it works ... the tab bar is _always_ full

even when you close a non-active tab with middle-mouse-click?

>I am going to revert this commit for now.

ok

eranif:
It seems so

kluete:
ok I did test your original code but I'll try again.

Given the logic in clTabCtrl::UpdateVisibleTabs()

  // don't update the list if we don't need to
  if(IsActiveTabInList(m_visibleTabs) && IsActiveTabVisible(m_visibleTabs)) return;

it seems that closing a non-active tab while the active tab is visible will prevent a refill.

If I can reproduce the bug do you want a step-by-step usecase or is there no point since visibility will depend on our screen resolutions & fonts?

-- p

PR #1039, opened 2015-11-25
DBE plugin fixes

MichalBliznak:
- Fixed missing ERD canvas in DBE plugin when built with wxWidgets 3.1.
- Fixed various build warnings in wxShapeFramework library.
- Fixed CMake script to allow build also on x86_64 version of Ubuntu 15.10.


eranif:
The fix for `CMakeLists.txt` should be `if (UNIX AND NOT APPLE)`

PR #1079, opened 2015-12-26
Handle Global include Path with Space Character

srini2174:
1) Fix to handle the global Include path with space character in it
2) Remove the trailing backslash character in DESTDIR variable.  This causes a wrong DESTDIR to be created by qmake in the Generated makefile.
3)  This fix has been tested in Qt 5.5.  I have not tested it in Qt 4.


srini2174:
I do not how to create a pull request only for specific patches/commits.  I have created this pull request for all the changes by mistake.  Can you please add only the commits "91533d8" "701f204" and "aa88fec" to the main repository.

eranif:
Thanks for the pull request, but in addition to the NMake build it contains some fixes that I won't apply:

* Update to `update64.bat` script 
* Fixes to `wx-config`

eranif:
Applied manually. FYI: to create a separate diff add the `.patch` suffix to the URL above, e.g: 
https://github.com/eranif/codelite/pull/1079.patch and edit it manually

PR #1080, opened 2015-12-26
Fix building for mingw with cmake.

Alexpux:
Fix building after replacing wxStandardPaths with clStandardPaths


Alexpux:
Can it be applied now?

Alexpux:
Ping?

Alexpux:
Thanks!

PR #1106, opened 2016-01-19
Added individual svg files for small resolutions of the CodeLite logo

DoctorRover:
Added individual svg files for small resolutions of the CodeLite logo.
Script create-logo.sh for automatic generation of png's from svg's is modified to account for the new svg files.
codelite-logo.ico file is updated to make it multi-image including a number of bitmaps with different sizes.


eranif:
Hi, I am going to revert this PR since it the low res images are not "clean" (you can see dark spots in the  edges of the squares) I guess that this is because that the "low-res" svg files were generated from the original SVG file. I have created a new low-res SVG file to generate clean images for 16/32 and 64 size

PR #1107, opened 2016-01-23
Line copy/cut feature

nghiaho12:
I miss the ability to copy and cut line, which is present in Geany and CodeBlocks.


eranif:
IIRC, `Ctrl-C` with no selection, copies the entire line
Doesn't this works for you?

nghiaho12:
Oh, I had no idea that even worked!

nghiaho12:
Cut line is still useful.

eranif:
Having a dedicated keyboard shortcut for this is a good idea, so we can keep both functionalities.
In fact, having this as a menu entry in the menu will expose people to this functionality

sausagejohnson:
@eranif  I'd like to revisit this issue if possible. Is it possible to simply have one keyboard shortcut that works like this for both situations:

* For control-X and there is nothing selected, the whole line will be cut to the clipboard
* For control-X and there is text selected, the selected text will be cut to the clipboard
* For control-C and there is nothing selected, the whole line will be copied to the clipboard
* For control-C and there is text selected, the selected text will be copiedto the clipboard

PR #1109, opened 2016-01-27
Faster tab navigation

nghiaho12:
Added to the View Menu
- Next tab
- Previous tab

I bind this to a keyboard shortcut for faster tab navigation than currently. I used this feature a lot in Geany.


echoma:
great commit and I am so glad to see eranif has merged this

PR #1246, opened 2016-04-18
Don't remove leading whitespaces in InsertExpansion

ilius:
User may want to add some spaces/tabs to beginning of abbrevation code as indentation, but `text` is being `Trim`ed because of extra indentation added by `FormatTextKeepIndent`
My change may seem a little complicated. But I didn't find a simpler solution (without adding a flag to `FormatTextKeepIndent`)


ilius:
Oh, and `text.Strip` is not supposed to be used in new codes due to wxString reference.

PR #1272, opened 2016-05-05
* Fixed MinGW GCC Compiler to use wxWidgets team built binary files.

stahta01:
I do NOT use CodeLite; but, I use Code::Blocks with wx-config. And, I added code to make wx-config to see the wxTeam built binaries tested the patch by compiling it with Code::Blocks.

I hope my patch works better than the one you reverted. Mine is smaller and only changes stuff I can test.

Tim S.

### DIFF
--- a/sdk/wxconfig/wx-config-win.cpp
+++ b/sdk/wxconfig/wx-config-win.cpp
@@ -890,6 +890,7 @@ public:
 
 
         // ### Variables: ###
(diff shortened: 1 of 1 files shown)


stahta01:
I just found the reason you reverted the other patch it fails to work correctly like mine does.

Please reject my change request.

Tim S.

eranif:
:)

PR #1286, opened 2016-05-23
Fixes and additions to quickfindbar

jgmdev:
When activating 'Highlight occurrences' the previous highlighted results where kep even when doing a new search so deactivating 'Highlight occurrences' was needed to remove them. This solves this issue by highlighting occurrences while one is typing.

This pull also adds a counter for the amount of highlighted occurrences which is good to know when doing micro refactoring on a single file using this tool. Overall this small changes should provide a better experience using the quickfindbar which I noticed on the Atom editor that works in a similar way.

### DIFF
--- a/LiteEditor/quickfindbar.cpp
+++ b/LiteEditor/quickfindbar.cpp
@@ -124,12 +124,12 @@ QuickFindBar::QuickFindBar(wxWindow* parent, wxWindowID id)
     m_regexOrWildButton->SetToolTip(_("Use regular expression"));
 
     // Marker button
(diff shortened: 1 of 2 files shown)


eranif:
Hi,
Thanks for this PR (looks great)
One question: why did you choose to display the "N results found" message on a separate panel? why not using the status bar for this?

jgmdev:
I thought having the results in the same panel is more visible for the given context, also some people (like me) may disable the status bar for some additional vertical space. 

By the way, while using this I noticed I didn't properly clear the marked lines if no results found will check where the change is needed and send another pull later.

Issue #1364, opened 2016-08-09
missing #include <unistd.h> for macOS in WordCompletion.cpp


dengxf2008:
when build in WordCompletion, got below error: 
fix it by add #include <unistd.h>

Undefined symbols for architecture x86_64:
  "isatty(int)", referenced from:
      word_init_buffer(yy_buffer_state*, __sFILE*, void*) in WordTokenizer.cpp.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [lib/WordCompletion.dylib] Error 1
make[1]: *** [WordCompletion/CMakeFiles/WordCompletion.dir/all] Error 2
make: *** [all] Error 2

eranif:
Building it

akien-mga:
For the reference, the PR mentioned above in #1366.

akien-mga:
Readding my comments here as #1366 was closed.

It also fails linking on Mageia 6 x86_64 (a GNU/Linux distro, i.e. the issue is not macOS-specific):
```
[ 68%] Building CXX object codelite_make/CMakeFiles/codelite-make.dir/cl_make_generator_app.cpp.o
cd /home/akien/Mageia/Checkout/codelite/BUILD/codelite-9.2/build/codelite_make && /bin/c++   -DGTK_USE_NATIVEBOOK=0 -DHAS_LIBCLANG_BRIEFCOMMENTS=1 -DINSTALL_DIR=\"/usr/share/codelite\" -DN$
CMakeFiles/WordCompletion.dir/WordTokenizer.cpp.o: In function `word_init_buffer(yy_buffer_state*, _IO_FILE*, void*)':
/home/akien/Mageia/Checkout/codelite/BUILD/codelite-9.2/build/WordCompletion/WordTokenizer.cpp:1599: undefined reference to `isatty(int)'
collect2: error: ld returned 1 exit status
WordCompletion/CMakeFiles/WordCompletion.dir/build.make:297: recipe for target 'lib/WordCompletion.so' failed
```

Note that 9.1 used to build fine.
Here is my RPM spec file: http://svnweb.mageia.org/packages/cauldron/codelite/current/SPECS/codelite.spec?view=markup
And the patches we're using (especially to unbundle external libraries): http://svnweb.mageia.org/packages/cauldron/codelite/current/SOURCES/

eranif:
Instead of adding the `#include <unistd.h>`
[...]

PR #1366, opened 2016-08-09
fix a link error on macos

dengxf2008:


eranif:
I just had a look at your patch and it can not be applied from several reasons:

* You added the missing(?) header in an auto generated file `WordTokenizer.cpp` which is generated by `WordTokenizer.l` (flex file) - this means that next time that I will re-generate it, you change will be lost
* Looking at `WordTokenizer.l` I specifically set `#define YY_NO_UNISTD` and it works on my MacBook... (OSX 10.11.5)

You can see at around line `521` in `WordTokenizer.cpp`:

```c++
#define YY_NO_UNISTD_H
#define YY_USER_ACTION  yycolumn += yyleng;

/* options */

#define INITIAL 0

#ifndef YY_NO_UNISTD_H
/* Special case for "unistd.h", since it is non-ANSI. We include it way
 * down here because we want the user's section 1 to have been scanned first.
 * The user has a chance to override it with an option.
 */
#include <unistd.h>
#endif
```

So I am really am puzzled about why it fails to link on your machine

akien-mga:
It also fails linking on Mageia 6 x86_64:
```
[ 68%] Building CXX object codelite_make/CMakeFiles/codelite-make.dir/cl_make_generator_app.cpp.o
cd /home/akien/Mageia/Checkout/codelite/BUILD/codelite-9.2/build/codelite_make && /bin/c++   -DGTK_USE_NATIVEBOOK=0 -DHAS_LIBCLANG_BRIEFCOMMENTS=1 -DINSTALL_DIR=\"/usr/share/codelite\" -DN$
CMakeFiles/WordCompletion.dir/WordTokenizer.cpp.o: In function `word_init_buffer(yy_buffer_state*, _IO_FILE*, void*)':
/home/akien/Mageia/Checkout/codelite/BUILD/codelite-9.2/build/WordCompletion/WordTokenizer.cpp:1599: undefined reference to `isatty(int)'
collect2: error: ld returned 1 exit status
WordCompletion/CMakeFiles/WordCompletion.dir/build.make:297: recipe for target 'lib/WordCompletion.so' failed
```

akien-mga:
Note that 9.1 used to build fine.
Here is my RPM spec file: http://svnweb.mageia.org/packages/cauldron/codelite/current/SPECS/codelite.spec?view=markup
And the patches we're using (especially to unbundle external libraries): http://svnweb.mageia.org/packages/cauldron/codelite/current/SOURCES/

eranif:
You have fixed the wrong file. `WordTokenizer.cpp` is an auto generated file generated from `WordTokenizer.l` file

You should fix this one instead. Also, please surround your code with 
```C++
#ifdef __WXOSX__
#endif
```

akien-mga:
Which obviously wouldn't fix the linking issue on Mageia, which is a GNU/Linux distro.

akien-mga:
Could you reopen #1364 maybe so that the bug can be discussed in the original issue, and not in an invalid PR?

PR #1378, opened 2016-08-14
MSVC improvements

MaartenBent:
I tried using the VC++15 compiler in CodeLite, but it was not detected automatically. So I added a generic MSVC compiler detector that detects past and future compilers.

Not all tools, compiler switches, paths, were correct. Improved this by executing `vcvarsall.bat` before `make`, which loads all the MSVC environment variables.

Also added 64bit compiler detection. Adding cross-compilers is also possible (platforms `x86_amd64` and `amd64_x86`).

And added some compiler and linker options.

Tested with vc10 and vc15.


MaartenBent:
Just noticed `VS90COMNTOOLS` will not be detected with [this ](https://github.com/eranif/codelite/pull/1378/commits/dec2e54751a99e898f2112fbfa8c93a76441b3d9#diff-d0297e391c7e82c09da83c3e30a05cd2R49) mask. I'll fix it later.

PR #1384, opened 2016-08-21
Fix PHPDoc bug

pattisahusiwa:
Wrong assign matches index when parsing @var type $name. This bug affected on assign wrong typehint value in variables_table.


eranif:
When submitting a PR, please avoid including changes to the `.workspace` file
(unless you added new project, which is very unlikely to happen)
I fixed this specific bug manually by taking the relevant part from the PR and merged it manually

PR #1393, opened 2016-08-27
Add some file extensions for default php project file types

anatooly:
Add php project file extensions for stylesheet preprocessor less, symfony framework yaml config, smarty template system.

*.yaml - use on my large project and symfony framework;
*.tpl - use on all Smarty template system;
*.less - use another css preprocessor format like scss;


eranif:
See my comment on Gitter

Issue #1100, opened 2016-01-13
[WebTools][PHP Workspace] JavaScript Find Definition not work


anatooly:
System: Linux anatooly 3.16.0-57-generic #77~14.04.1-Ubuntu SMP Thu Dec 17 23:20:00 UTC 2015 x86_64 x86_64 x86_64 GNU/Linux
Workspace: PHP
Example file test.js:
```
function fooTestName() {
	return true;
}

function barTestName() {
	return false;
}

function bazTestName() {
	return 1;
}

barTestName();
```

1) Create this file on Node.JS workspace - 'Find Definition' work correct (function select);
2) Create this file on PHP workspace - 'Find Definition' not work (function not select);

Log (first is Node.JS, last is PHP Workspace):
```
[ 11:41:29:057 DBG ] ContinuousBuild::OnFileSaved
[ 11:41:29:057 DBG ] ContinuousBuild is disabled
[ 11:41:36:341 DBG ] [WebTools] POST / HTTP/1.1
From: Eran Ifrah
User-Agent: CodeLite IDE
Content-Type: application/x-www-form-urlencoded
Content-Length: 328

{
 "query": {
  "type": "definition",
  "file": "#0",
  "end": {
   "line": 12,
   "ch": 6
  }
 },
 "files": [{
   "type": "full",
   "name": "test.js",
   "text": "function fooTestName() {\n\treturn true;\n}\n\nfunction barTestName() {\n\treturn false;\n}\n\nfunction bazTestName() {\n\treturn 1;\n}\n\nbarTestName();\n"
  }]
}
[ 11:41:36:421 DBG ] {"origin":"test.js","start":51,"end":62,"file":"test.js","contextOffset":50,"context":"unction fooTestName() {\n\treturn true;\n}\n\nfunction "}
0
[ 11:43:00:915 DBG ] [WebTools] POST / HTTP/1.1
From: Eran Ifrah
User-Agent: CodeLite IDE
Content-Type: application/x-www-form-urlencoded
Content-Length: 328

{
 "query": {
  "type": "definition",
  "file": "#0",
  "end": {
   "line": 12,
   "ch": 4
  }
 },
 "files": [{
   "type": "full",
   "name": "test.js",
   "text": "function fooTestName() {\n\treturn true;\n}\n\nfunction barTestName() {\n\treturn false;\n}\n\nfunction bazTestName() {\n\treturn 1;\n}\n\nbarTestName();\n"
  }]
}
[...]

PR #1398, opened 2016-09-01
Fix Find definition for JS files on PHPWorkspace & CppWorkspace

anatooly:
This fix on bug: https://github.com/eranif/codelite/issues/1100


eranif:
This fix is wrong a bit.
By default, `ternjs` will return the location of the definition in relative path.
You patch assumes that it is relative to the current file path. However, it is relative to the location of the `.tern-project` file. For NodeJS workspace, this file is created under the root folder (where the workspace is located) and for other workspaces, this file is created under `WORKSPACE-PATH/.codelite` folder

eranif:
@anatooly I just committed: https://github.com/eranif/codelite/commit/f8db4f3561386a4ee6bf182638342114f2ff30b2 which uses a simpler approach and addresses the same fix you attempted to fix here.

On a side note:
 I **really** hope that I don't discourage you from sending pull requests :/

anatooly:
Thanks a lot.

PR #1425, opened 2016-09-23
Fix several tab issues

pattisahusiwa:
This PR fixed tab size on left or right orientation and light color on dark theme.

issue

fix

### DIFF
--- a/CodeLite/PHPLookupTable.h
+++ b/CodeLite/PHPLookupTable.h
@@ -301,7 +301,7 @@ void PHPLookupTable::RecreateSymbolsDatabase(const wxArrayString& files,
 
         m_db.Begin();
         for(size_t i = 0; i < files.GetCount(); ++i) {
-            if(pFuncGoingDown && pFuncGoingDown()) {
+            if(pFuncGoingDown()) {
                 break;
             }
             {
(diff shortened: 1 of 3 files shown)


eranif:
On Windows, this patch breaks the tab rendering completely - i.e. the tabs are drawn all in system background colour and are completely invisible.

I restored the tab size to use 3 pixels on non Windows systems and 5 for Windows.

eranif:
Just committed another fix that changes the logic for how to check if a colour is considered dark.
Following this answer: http://stackoverflow.com/questions/12043187/how-to-check-if-hex-color-is-too-black
Please check and see if it helps you

pattisahusiwa:
Never mind. Your link is used this formula [https://en.wikipedia.org/wiki/Grayscale](https://en.wikipedia.org/wiki/Grayscale). It's up to you to set threshold value for Ylinear. But, by using your method, `#292929` and above will be considered as light. When using my method, light will start at `#808080`. For me, `#292929` is NOT light.

pattisahusiwa:
This line won't compile on my machine `if(pFuncGoingDown && pFuncGoingDown()) {`. Please fix it.

eranif:
I wondered why you removed it...
Will fix this

eranif:
Done (compilation error)

Issue #1433, opened 2016-09-28
PHP Parser not Working with Special Characters


jgmdev:
Adding special characters to comments like áéíóúäëïöü, etc... breaks the PHP Parser resulting in a empty outline and code completion symbols database. This worked before so it seems that a recent commit broke it.

Here is some sample code that previously worked and now doesn't (notice the **á**):

```php
<?php
/**
 * Something
 * @author First Lást Name
 */

 class MyClass
 {
     private $var1;

     private $var2;

     public function functionOne(){}

     public function functionTwo(){}
 }
```
but if I remove the **á** then it works:

```php
<?php
/**
 * Something
 * @author First Last Name
 */

 class MyClass
 {
     private $var1;

     private $var2;

     public function functionOne(){}

     public function functionTwo(){}
 }
```

eranif:
Seems to be working here...
<img width="569" alt="works" src="https://cloud.githubusercontent.com/assets/3677711/18932931/263bb626-85db-11e6-93a5-7c5903132f90.png">

jgmdev:
thats strange... I just built the latest code from github (ArchLinux) using https://aur.archlinux.org/packages/codelite-git (which I maintain) and it isn't working...

jgmdev:
Here is a screenshot of mine
![test](https://cloud.githubusercontent.com/assets/1702572/18933471/f6321700-85a2-11e6-82fc-c3b9c7c7d59c.png)

eranif:
There is a bug that if the caret is the last char, the completion is not working. add some empty lines at the end of the file and try to code complete while making sure that the caret is *not* at the last position

jgmdev:
Doesn't seems to be the case, I even compiled wxWidgets with gtk3 and various different flags but it didn't worked, here some screenshots that show the version with special character and non special character

![acute](https://cloud.githubusercontent.com/assets/1702572/18943919/83b206ba-85f0-11e6-8054-0e9826b908b0.png)

no special character

![non-acute](https://cloud.githubusercontent.com/assets/1702572/18943925/8dc404dc-85f0-11e6-938c-672d5b575ba2.png)

Also, I compiled the latest stable version 9.2.0 and that one works.

eranif:
Works on Debian 8 as well :/ is it possible to get a minimal workspace that mimic the problem? (maybe its not that single file that causes the problem)

pattisahusiwa:
[...]

PR #1434, opened 2016-09-29
Workaround for issue #1433.

jgmdev:
Here is a pull request for issue #1433  for easier review and testing.


eranif:
Please check my latest commit which fixes this in a generic way without the need of the workaround

PR #1505, opened 2017-01-07
surround some PCH stuff with if(APPLE)

ysblokje:
This seems to be done before at multiple places.


ysblokje:
Oh this enables compiling codelite with ninja build on linux at least.

PR #1510, opened 2017-01-10
Add option to use dark tab theme by default

hdastwb:
CodeLite's dark tab theme looks really good when used with other dark-themed UI elements, but there isn't currently a good way to configure CodeLite to use that theme (besides the match-editor-theme option, which only works for editor tabs). These changes add a new option to the settings dialog (Windows & Tabs > Docking > Use dark notebook tab theme by default) which allows the user to switch the tabs to the dark theme (when not overridden by the match-editor-theme option). This option is stored using what appears to be a previously unused options bit, and should have all of the logic in place for initializing and updating the color theme of all four of CodeLite's main notebooks.

I've never used wxCrafter before making this small change, and, since it looks like it clobbered pretty much all of the .wxcb file and the related #include guard, I'm worried that I might be doing something wrong or using the wrong version. I can fix those files manually if you want me to.

I'm also not sure if there's anything special that I need to do to get the new text that's added with this change to be localization-ready; if there is, I'll take care of that too.

### DIFF
--- a/LiteEditor/debuggerpane.cpp
+++ b/LiteEditor/debuggerpane.cpp
@@ -94,6 +94,10 @@ void DebuggerPane::CreateGUIControls()
     if(!EditorConfigST::Get()->GetOptions()->GetWorkspaceTabsDirection()) {
         style |= kNotebook_BottomTabs;
     }
+    if (EditorConfigST::Get()->GetOptions()->IsTabColourDark()) {
+        style &= ~kNotebook_LightTabs;
+        style |= kNotebook_DarkTabs;
+    }
     style |= kNotebook_UnderlineActiveTab;
 
     GeneralImages img;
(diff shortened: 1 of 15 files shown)


eranif:
I applied the patch locally and this look OK.
I wonder which version of wxCrafter are you using since I noticed that the `wxcp` file contains TABS.
While latest versions of wxCrafter are using SPACES (2 chars) to indent which makes the file more readable (to me at least).

Is it also possible to support the following notebooks:

* NodeJS debugger notebook see class `NodeJSDebuggerPane`
* PHP Debugger pane notebook see class `PHPDebugPane`

hdastwb:
I just grabbed the evaluation version off of the website for Ubuntu 16.04; it looks like that puts me on version 2.6.

In any case, I fixed the wxCrafter inconsistencies introduced earlier and updated `NodeJSDebuggerPane` and `PHPDebugPane` (along with `EvalPane` too) so that they should all follow the new setting now.

PR #1639, opened 2017-05-21
Phpcs (RFC)

AJenbo:
Hi, I started working on some extensions for CodeLite, I have gotten it to a POC level, but I'm unsure of what direction to take from here, so hope a pull request is an okay way to request for comment an possibly a little help. 

TODO:
- Figure out if this should be part of the php/codeformatter-plugin or stays as a separate plugin (I also plan to implement phpmd (and posibly phpcpd) integration)
- Figure out how to integrate with config from other plugins. Useful for getting phpexe, but also to do a proper fix for phpcs.cpp:106
- Figure out how to create a settings modal (I have never done GUI in c++, and I'm still a novice at c++)


eranif:
This pull request has some conflicts. Can you please fix this?

AJenbo:
Sure. I but merging it would probably be a bit premature unless you feel like making the gui/conf part of the plugin :)
Also I added a quick implementation (the path for the project config.xml is hardcoded and i plan to use XML instead of text output for better parsing and options) of a phpmd as I really needed it for work.

eranif:
If you are having problems creating the GUI parts, I can create the skeleton configuration dialog and you can take it from there (I am using `wxCrafter` to create the UI, also part of CodeLite)

AJenbo:
@eranif that would be much appreciated. All I really need is a couple of input boxes for giving the path to php, the linter (phpcs/phpmd) and in phpmd's case a rules.xml.

I just did the XML parsing for the phpmd plugin so now the GUI is all that I need for this to be ready :)

eranif:
So, you need a simple settings that can select 3 files, right?

AJenbo:
I haven't figure out how i'm supose to open wxcrafter in CodeLite, so instead I have installed the standalone version. At first it failed to launch (missing libpng12, im on Ubuntu 17.04), but after downloading the missing lib file it seams to work. I'll se if i can figure out how to get this working. I just saw a commit for the vim plugin that looks like it could be helpful in figuring out how to do this. I'll let you know how it goes :)

eranif:
Better working from within CodeLite.
To add new dialog, Right click on a folder (the yellow ones) and select `wxCrafter->Add wxWidgets UI form`

A wizard will come up...

1. 1st page, select wxDialog and give it a title
2. 2nd page, click on the button next to the drop down control (white with green plus sign) to create a new form and dont re-use an existing one. Next you will need to setup the base name. Pick something like `PHPCSSettingsDlgBase` Hit TAB to move to the next field and you will notice that wxCrafter will suggest the subclass name and file to be the same as the base name class just without the `Base` part
3. Select the folder where to place the sources

click Finish and you are done. You can now edit the UI. 

See this tutorial: http://codelite.org/LiteEditor/WxCrafterHelloWorld

AJenbo:
I only get an option for wxFormBuilder :/ I'm also missing the wxCrafter tab from the tutorial

mtangoo:
>I only get an option for wxFormBuilder :/ I'm also missing the wxCrafter tab from the tutorial

Is it listed in Plugins? You seem to use Ubuntu, did you install wxCrafter also? It works flawlessly in my Ubuntu following the guide on the website and then install with `sudo apt-get install codelite wxcrafter`

eranif:
Which OS are you using? it should be bundled with CodeLite (wxCrafter plugin)
Is it a self compiled CodeLite or an official version?

AJenbo:
Ubuntu 16.04 and also on Ubuntu 17.04, currently I'm using a self compiled version (the one in this pull request) but I can try the official on and the distribution later.

AJenbo:
The plugins is listed and enabled

mtangoo:
>I'm using a self compiled version (the one in this pull request)  

I have never tried that myself, so cannot help. But installing official  will save you pain on this.

AJenbo:
Ok I have re-factored the linters in to a single plugin and the code for that part is pretty much what I would consider ready, unless you have some comments.

As for the options dialog I managed to create something in wxCrafter that looks right, but when I try to open the dialog from the menu it just crashes CodeLite instantly (no error in the logs) and I for one can not figure out how to get it to work. Any help at this point would be greatly appreciated.

eranif:
This sounds like missing symbols issues.
To fix this: you need to re-run `cmake` and build again so the new generated files will get included and linked

AJenbo:
Yep that fixed it :(
(sad because i spent hours trying to figure out what I did wrong, redoing it a few times)

AJenbo:
@eranif I'm happy to say that I now think the plugin is ready for merge. Hope you feel the same :)

eranif:
Great, I will merge it later today and will test it on Windows (I am not sure if you tried it on Windows)

AJenbo:
Thanks that would probably be a good idea, though I have strived to keep it compatible.

I have yet to figure out how to compile CodeLite on Windows. The README.md skips setting up wxWidget and the download link is doesn't work (looks like the right package is DwxMSW-3.0.3_gcc510TDM_x64_Dev.7z), but i still end up with a non informative fail about halfway threw the compile (current master branch on your repo mind you).

eranif:
FYI: CodeLite is built with wxWidgets 3.1 on Windows and it uses some custom build steps which are explained here in details:

http://codelite.org/Developers/BuildingWxWidgetsWin

eranif:
Are the tooltips in the dialog correct? (e.g. the field that says: `PHPMD`, the tooltip is: "Select the PHP-CS-Fixer phar file location")
Can you please go over the dialog entries and correct the tooltips?

eranif:
Few comments:

- Avoid using the old macros `CL_DEBUG`, `CL_WARNING` etc, please use the new macros. An example of a typical debug line: 
```C++
clDEBUG() << "The filename is:" << m_fileName << clEndl;
```
- I fixed a bug that if no PHP exectuable was set properly and then I fixed it from the PHP settings dialog, it was not reloaded by the PHPLint plugin

other than that, I added some more debug messages to the log and ensured that the plugin is built properly on Windows (and installs)

AJenbo:
Thanks for the help :)

Would it be helpful if I cleaned up other usages of CL_DEBUG, CL_WARNING & frinds?

I'll keep the Windows install in mined if I do another plug-in. The only major feature I have plans for is PHP refactoring, but I'm not sure weather i should use an external phar tool or work with the build in lexer (might be above my skill set, but if you feel like helping I could probably learn).

AJenbo:
P.s. i think PHPLint and the VIM mode plugins are missing from the list in the initial setup wizard. I looked threw the code but came to the conclusion that it's generated dynamically, but didn't really debug the issue.

AJenbo:
After setting following the guide you linked to mingw32-make reports that there isn't a file called Makefile.gcc in the project.

Running build crom codelite now gives me this more informative error:
windows-error

mingw32-make with Makefile instead of Makefile.gcc resultes in an error about missing SDK files for sqllite3 :/

eranif:
This means that you did not build wxWidgets properly. The `wxTrap` calls are part of the Debug build and they should not be called when building CodeLite in release mode.

I will run my guide on a clean setup, step by step to make sure I did not miss anything there.

About the plugins missing in the wizard: I do see them... Just run "Help->Run the setup wizard" - and I can see them there

PR #1657, opened 2017-06-01
Locate dependencies on unix via which

AJenbo:
php, clang_format????, php-cs-fixer/phpbcf, phpcs and phpmd are now found
automatically (using which command) if installed on the system.

The autodetect is also moved to the loading of the config to avoid
having to restart CL for it to autodetect again if values are emptied,
giving the user more immediate feedback.


eranif:
instead of using an external command (e.g. `which`) which does not exists on Windows, I would rather mimic the `which` command behavior in a class, or a global functions, something like:

EDIT:
I just committed such a function, please use it instead of executing an external command, in addition it can be used on all platforms

See this commit: https://github.com/eranif/codelite/commit/66c088232c8b6061f06ecbe308010cb0aa244641

AJenbo:
Cool, this means the PHPLocator could be completely striped as Getini() and Getincludes() weren't wired up to anything.

It might be a good idea to test that codelite-clang-format is correctly detected on non-__WXGTK__ systems.

I have still hidden autodetection of phar-files behind an #ifdef since it won't work on Windows. But that might be premature optimization?

eranif:
`codelite-clang-format` on Windows and OSX, is part of the installation - so there is no need to locate them, CodeLite knows where they are located

eranif:
I just noticed you removed the `codelite-clang-foramt` code for Windows/OSX - you should restore it

CodeFormatter/clClangFormatLocator.cpp
```diff
@@ -25,22 +24,14 @@ bool clClangFormatLocator::Locate(wxString& clangFormat)
...
-        if(fnClangFormat.Exists()) {
-            clangFormat = fnClangFormat.GetFullPath();
+        if (clFindExecutable(nameOptions.Item(i), file)) {
+            clangFormat = file.GetFullPath();
             return true;
         }
     }
     return false;
-#else
-    wxFileName fnClangFormat(clStandardPaths::Get().GetBinaryFullPath("codelite-clang-format"));
```
eranif: This part should be restored. on Windows/OSX `codelite-clang-foramt` is always located under `clStandardPaths::Get().GetBinaryFullPath`

AJenbo:
Ok, I was unsure how it tied in to things.

AJenbo:
Detection of phpcs/phpmd/phpcbf/php-cs-fixer should work on all Unix systems as phar is seen an exec format there. So since WXGTK is Linux only; what should I use instead to expose code to both Linux and Mac OS X (Unix systems)?

eranif:
You can either use:
```C++
#ifndef __WXMSW__
#endif
```

Or:

```C++
#if defined(__WXOSX__) || defined(__WXGTK__)
#endif
```

FYI: this PR can not be merged - it has conflicts

AJenbo:
Thanks. Conflict resolved.

CodeFormatter/phpoptions.cpp
```diff
@@ -28,6 +34,5 @@ PhpOptions& PhpOptions::Load()
 
 PhpOptions& PhpOptions::Save()
 {
-    clConfig config("php.conf");
```
eranif: I think that the actual bug here is this:
```C++
PhpOptions& PhpOptions::Save()
{
    clConfig config("php.conf");
    config.WriteItem(*this);
    return *this;
}
```

PHPLint/phpoptions.cpp
```diff
@@ -0,0 +1,38 @@
...
+{
+    clConfig config("php.conf");
+    config.ReadItem(this);
+    return *this;
+}
+
+PhpOptions& PhpOptions::Save()
+{
+    return *this;
+}
```
eranif: Same thing here: `Save` does nothing...
it should do this:

```C++
clConfig config("php.conf");
config.WriteItem(this);
return *this;
```

Also, why do use the same file name `php.conf` !? it will override the default php settings

eranif: Also, avoid name two classes with the same name `PhpOptions` (unless I am missing something here, there is one under `PhpLint` folder and one under `CodeFormatter` - this will definitely cause a confusion the C++ code completion parser.
Please rename it

CodeFormatter/phpoptions.cpp
AJenbo: The phpoptions.cpp is there to allow the plugins to use chare the phpexec configuration so that the user only has to set it up once instead of in 3 different places to make each plugin work.

The plugins have there own config file that handels there specific settings.

PHPLint/phpoptions.cpp
AJenbo: The save function is a stub exactly to avoid the chance of it overwriting php.conf.

I tried simply including php_configuration_data.h and adding "${CL_SRC_ROOT}/codelitephp/PHPParser" as include path in the cmake files. This does compile, but at the end the application crashes with the following error:

[22:55:44:466 ERR] Failed to create single instance socket: CreateServer: bind operation failed: Address already in use
codelite: symbol lookup error: /usr/lib/x86_64-linux-gnu/codelite/CodeFormatter.so: undefined symbol: _ZN20PHPConfigurationDataC1Ev

Am I missing something or should it be solved in another way?

eranif:
Sorry to insist on this, but creating dependency between two plugins is against the general design of CodLite. Each plugin should be able to compile and load without a dependency of another plugin.
If you think that `PhpOptions` (or any other class for that matters) should be shared across plugins, then move this class to the `Plugin` folder (project `plugin_sdk` in CodeLite's workspace) 

Note that classes created under the `plugin_sdk` project should be using the macro `WXDLLIMPEXP_SDK`, for example:

```C++
#include "codelite_exports.h" // needed for the macro WXDLLIMPEXP_SDK 
class WXDLLIMPEXP_SDK clFindResultsStyler : public wxEvtHandler
{
};
```

eranif:
Ok, this is what I did:

- Moved the class `PhpOptions` from `CodeFormatter` to `Plugin` project, this means you can safely include it in every plugin
- Fixed the `Save()` method to actually save
- Added new CodeLite event named `wxEVT_PHP_SETTINGS_CHANGED` , this event is sent when the `PhpOptions` are saved to the disk. So other objects are able to reload the settings, in your code, it should be something like this:

```C++
EventNotifier::Get()->Bind(wxEVT_PHP_SETTINGS_CHANGED, &MyClass::OnPhpSettingsChanged, this);
...
void MyClass::OnPhpSettingsChanged(clCommandEvent& event)
{
    event.Skip(); // Always call this, so other handlers will also get to process this event
    m_settings.Load(); // Reload the settings
}

```

eranif:
Ignore what I wrote earlier.. I moved the class `PhpOptions` to the `Plugin` folder, in addition I removed the basic PHP settings (exe path, include paths and error reporting string) and moved them to the new `PhpOptions` class (I also added migration code to copy the settings from the old file to the new file)

Commit: https://github.com/eranif/codelite/commit/8fcaacca93db3bb0714e26ff981133d9b9f619eb

AJenbo:
Thanks that was really helpful

eranif:
I have merged this PR, note that it had two build errors:

1. missing include header which I added
2. Undeclared identifier `m_flags` in `lintoptions.cpp` (I had to comment it until you fix it), see this commit: https://github.com/eranif/codelite/commit/14960778e12f01a57e441ba49aa64b3552885159

PR #1658, opened 2017-06-02
Add MemCheck to the list of C++ plugins for the welcome wizard

AJenbo:


eranif:
Can you surround this with `#ifdef __WXGTK__` (the MemCheck plugin is for Linux only)

AJenbo:
Done

PR #1663, opened 2017-06-05
Php refactoring browser

AJenbo:
As mentioned in the forum I recoment using my branch of the re-factoring tool as it has support for php7 and generally handles php syntax better (if it fails nothing is changed):
https://github.com/AJenbo/php-refactoring-browser


PHPRefactoring/phprefactoring.cpp
```diff
@@ -0,0 +1,306 @@
...
+        ::wxMessageBox(_("Can not refactor file:\nFailed to write temporary file"), "PHP Refactoring",
+            wxICON_ERROR | wxOK | wxCENTER);
+        return;
+    }
+
+    // Ensure that the temporary file is deleted once we are done with it
+    FileUtils::Deleter fd(tmpfile);
+
+    ::WrapWithQuotes(tmpfile);
+    parameters = action + " " + tmpfile + " " + extraParameters + " | patch -p1";
```
AJenbo: I make use of patch to apply the changes to the temp php file. I'm not sure how well this will work on Windows so probably it would be better to return the output and do the patching in the CL code. I noticed that CL already uses the dtl diff libery that might provide the needed functionality, but I'm unsure how to use it.

Not that for the refactoring that changes Namespaces to be PSR-0 complient (fix-class-names) the patch can contain changes to multiple files.

eranif:
You added the refactoring menu in the main menu-bar.
This is a bit inconvenient,  I would expect the following workflow:
- Right click in the editor on variable/word
- Get the PHP context menu with various options, among them: Sub-menu with the title `Refactor` which contain the various refactoring options

What do you think?

eranif:
Also, is it possible to write some docs on all the work you did (PHPMD, PHPCS, refactoring etc) on the wiki?

AJenbo:
Hmm I used the Copyright conext-menu and php-code-generation code as a starting point for mine. For me it does appear in the context menu when I right click the code I want to redactor. So yeah that’s how it's suppose to work. But I'll have a look at the other code and see if there’s a more optimal way I could do it.

billede

In regards to documentation my plan is to make the settings UI a bit more intuitive first (like the Code Formatter has links to the external tools in it's help section), but after that I'll defiantly write some documentation on the wiki for the work that I have done.

eranif:
Sorry, I missed... that you also used `OnEditorContextMenu` event handler ... this is what I meant, and you seems to done it properly :)

AJenbo:
Looks like I might need some additional permissions to edit the Wiki. Should I create them as part of the php page or seporate pages for each plugin?

eranif:
Send me your email, I will send you a user name and password for the wiki.
And I prefer to have them as part of the same PHP section if thats OK

AJenbo:
Any advice on how to proceeder with this one :) As for documentation I think I might have some time over the weekend.

eranif:
Is it in a working condition? I can merge it (+test and compile on OSX and Windows) if this is what you are waiting for
Let me know

AJenbo:
The fix I just pushed gets it to where it's fully working on platforms that can pipe to patch. But I don't think that includes Windows (see my comment in PHPRefactoring/phprefactoring.cpp). So if you know how to use dtl diff (or some other library) to achieve this I would make the thing more portable. Other then that it's been working nicely for me for the past few days :)

AJenbo:
It looks like a version of patch.exe is also included with CL, so possibly making use of it would work. Even better would be if we could parse the unified diff internally and apply it to the files/editor so that only the actual changed lines will show as changed rather then the whole file (as is also the case for the code formatter).
So yeah, ideas/opinions on how to apply unifed-patch in a OS agnostic way is what i need :)

eranif:
Maybe we could split this into two commands? the first to generate the patch file and the second applies it?

eranif:
I just checked the Windows installation file and it looks like (as I remembered) `patch.exe` is included so we can assume that we have `patch` utility

AJenbo:
Sounds reasonable, most solutions would require that the patch part be separated anyways.

Would we need a config option for setting the path to patch or can we assume it is auto discoverable on all platforms?

eranif:
I added new `clPatch` class (which uses `patch.exe`) - you can use it

eranif:
Notice that the class `clPatch` throws an instance of `clException` incase of a failure

eranif:
I just noticed (on the forum) that you are using your own forked version of the refactoring tool.
So I wonder: maybe for solving the patch issues - can you generate the modified files (without diff) and by doing this we could use DTL library to merge the two files without using an external tool.

e.g., lets assume you want to modify the file: `class.php`, instead of generating a patch file (or diff), create a file named `refactored-class.php`  (you can either place it next to the original unmodified file, or in a separate folder) DTL can then be used to merge the two files (i.e. apply `refactored-class.php` onto `class.php`). Just an idea

AJenbo:
I would prefer to keep the fork as compatible with the original as possible. Even though the project is dormant at the moment I hope that the code can be upstreamed at some point. But maybe it would be ok to add as an optional parameter. Also when I initially looked in to it, it seamed that it is generating the diff directly so rewriting it to output a full file might be a bit of work.

But it's time that I would rather spend implementing php-refactoring natively in CL.

eranif:
> But it's time that I would rather spend implementing php-refactoring natively in CL.

Does this mean that this PR is no longer required?

AJenbo:
No, i think that would be in a fare distant future and at this point i don't even know if I'm capable of doing it.

This week is pretty packed for me, but next week I expect to make the last needed changes for this to be ready to be merged.

eranif:
Do we still need `patch.exe` on Windows for this to work?

AJenbo:
It relies on clPatch and as I understand it that relies on patch.exe?

eranif:
Yes, `clPatch` uses `patch.exe`. I will see if I can upgrade Windows `patch.exe` for next release

AJenbo:
I guess it's time to write some instructions on the wiki :)

eranif:
Will test this on Windows later on today ;) (installation wise, and functionality ofc)

eranif:
I just tried it on Linux:

- I wasn't able to rename a method name ( I tried 'rename class property') - is this supported?
- I tried renaming a local variable, it also failed. I did notice that error coming back from the `refactor` executable are not reported back to the user

Debugging it a bit revealed that the new local variable name must not start with `$` - I fixed that
Also note that `clPatch::Patch` throws an exception, so I wrapped it with `try & catch` blocks

eranif:
In case of a successful refactor, I added a call to `EventNotifier::Get()->PostReloadExternallyModifiedEvent(false);` which is supposed to force file reload from the disk (it makes sense cause you just modified it...)

AJenbo:
Files are not directly modified, a temp file is created, the change is then copied back in to the editor so that you can evaluate it perform an undo with out having saved to the file.

eranif:
Ok, I removed the call to `EventNotifier::Get()->PostReloadExternallyModifiedEvent(false);`
I sure hope that the `-p1 <` (the patch code) will work on Windows :)

AJenbo:
It should, I took it from a windows guide :)

I added two section to the php section of the wiki.
http://codelite.org/LiteEditor/PHP#toc7

Hope it describes the process well enough.

Some pictures might have been nice, but I'm not sure how to upload them.

eranif:
Thanks, I have read the documentation and fixed some typos .
Please read it and confirm that I did not make a mistake :)

AJenbo:
Thanks, you had the command download the orginial QafooLabs version, but my original instruction where a bit messy. I fixed it, tested it, and cleaned up the instructions a bit more :)

PR #1665, opened 2017-06-05
Codeformatter

AJenbo:
This fixes and issue where events and status messages where not always sent.

JavaScript is now formatted even if Clang is not the preferred C++ formatter.

Java is now also formatted using Clang-format.

Fixes a missing undo action for AStyle formatting.

Generally more consistent behaviour.


AJenbo:
Managed to get at bit more ready in this series:
phpcbf support.
Preview for php-cs-fixer.
Add options for php-cs-fixer.

There is still a lot more options to add, but the preview and additional tool support in it self should justify the change I think.

eranif:
I merged this, although I did notice a bug (which I already fixed): you don't load the `FormatOptions` moer than once, instead you keep a single instance of it and pass it to the `CodeFormatterDlg` when needed. The settings changed in the dialog are being applied to the dialog member named `m_options` (which is a different instance from the one kept in `CodeFormatter` class, so any changes are not reflected until a restart of CodeLite is done (so the options are reloaded from the file system)

Changing `CodeFormatterDlg` member from: `FormatOptions m_options;` to `FormatOptions& m_options;` (notice the ampersand) fixes this issue (I also changed its initialization, since reference can only be initialized in the constructor)

AJenbo:
Thanks, I actually noticed that issue today at work, but hadn't gotten around to debugging it. What really confused me was that it seamed to work for some of the tools, but I think that was because they get there parameters in a different manor.

There is still a lot of clean up to be done for this plugin as currently not all formatters are available for all functions.

Another thing I noticed is that AStyle src is included twise, once in the SDK/AStyle folder and once directly in the codeformatter plugin folder. Would it be simple enough to remove the extra copy from the plugin folder, or is there something special about how it is included here?

eranif:
I am not sure why, will need to check this, but to keep things safe, just leave it there :)

PR #1681, opened 2017-06-25
Codeformatter refactoring

AJenbo:
All formatter engines are now fully hooked up (able to do batch formatting and formatting events).

Code specifically relating to each tool has been reduced, probably a good next step would be to put them each in there own object as it would make things much more manageable.

The php sample has been put in a separate file.

The added 31 options for php-cs-fixer are now implemented.

Fixed a bug where settings would take affect for the current session even if pressing cancel.

Fix a typo in imanager.h

Preview is now workspace aware (clang_format, phpcbf, php-cs-fixer).

When formatting selection with clang only the formatted part will now show as updated.

Lines no longer shows as being updated if nothing was changed during selection formatting.


CodeFormatter/codeformatter.h
```diff
@@ -30,94 +30,74 @@
...
+enum CodeFormatterEngine {
+    cFormatEngineNone,
+    cFormatEngineAStyle,
+    cFormatEngineClangFormat,
+    cFormatEngineBuildInPhp,
+    cFormatEnginePhpCsFixer,
+    cFormatEnginePhpcbf,
+    cFormatEngineWxXmlDocument,
+};
+
```
eranif: In all the new code of CodeLite, enumerators starts with `k`, so it should be `kFormatEngineNone`, `kFormatEngineAStyle` etc

eranif:
Is this PR replaces the other one opened here? or should I apply this one on top of your previous one?

CodeFormatter/formatoptions.h
```diff
@@ -112,6 +112,16 @@ enum PHPFixserFormatterStyle {
...
+    kPcfCombineConsecutiveUnsets = (1 << 22),
+    kPcfLinebreakAfterOpeningTag = (1 << 23),
+    kPcfMbStrFunctions = (1 << 24),
+    kPcfNoBlankLinesBeforeNamespace = (1 << 25),
+    kPcfNoMultilineWhitespaceBeforeSemicolons = (1 << 26),
+    kPcfNoNullPropertyInitialization = (1 << 27),
+    kPcfNoPhp4Constructor = (1 << 28),
+    kPcfNoShortEchoTag = (1 << 29),
+    kPcfNoUnreachableDefaultArgumentValue = (1 << 30),
+    kPcfNoUselessElse = (1 << 31),
```
AJenbo: I'm running in to an issue here. enum has a limit of 32 items, but i will probably need about 250 to support all options. One thing I could do is to only split it up in to an enum per widget, but even like that i think there is a chance that i will have 60 options in one widget. Would you suggest just splitting it up in multiple widgets as well or is there another way to handle this?

AJenbo:
This PR is about improving the CodeFormatter-plugin, my other PR adds a new plugin for php-refactoring. So they are not related at all.

CodeFormatter/formatoptions.h
eranif: split them.
You can use up 64 if you use macros instead of `enum`, like this:

```C++
#define kOne (1ULL << 0)
#define kTwo (1ULL << 1)
#define kThree (1ULL << 2)
#define kLast (1ULL << 33)
```

eranif: Actually, when I think of it, we are using `C++11`, this means that you can use `class enum` like this:

```C++
enum class eEnum : unsigned long long {
    kOne = (1ULL << 0),
    kTwo = (1ULL << 1),
    kThree = (1ULL << 2),
    kLast = (1ULL << 33),
};
```
this allows you to set the type of the `enum` to `unsigned long long`
Note: whith class enum you need to reference it like this:
`eEnum::kOne`

AJenbo: Hmm while this works nicely for expanding the enum limit, it then results in another issue, setting the options in the dialog failes to compile as wxVariant isn't able to handel the 64bit int.

I'll work on splitting the config up. Maybe I can make a seporate config object for php-cs-fixer, it would make sense to save in the json config format as that is what php-cs-fixer takes as an input and so i would be able to reuse more code instead of all this format range-ling.

Of cause this would break comparability with the options that are currently in this PR, so if you prefere not to have a break like that you can wait with merging till I have the new format down.

eranif:
I'll wait then

PR #1685, opened 2017-07-01
Format PHPLint-plugin to the standard set in .clang-format

AJenbo:
There are only style changes here. If you prefer I can commit clean ups of this type directly in the future.
Also how would you feel about me cleaning up the style in code I haven't actually worked on in the past?

### DIFF
--- a/PHPLint/lintoptions.cpp
+++ b/PHPLint/lintoptions.cpp
@@ -12,17 +12,17 @@ LintOptions::LintOptions()
     , m_phpmdPhar("")
     , m_phpmdRules("")
 {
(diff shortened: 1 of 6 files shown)


eranif:
I am sorry, I did a mistake with the new `.clang-format`
* It failed to work on older `clang-format` command line tools (like the one I am using on Linux)
* It added an extra space  before control statements (`if` etc)

I have updated `.clang-format` file in the git repo

AJenbo:
No problem, I though there was a good chance that it would reveal some less then ideal settings :)
I opened a new PR where I used the updated .clang-format. My clang-format is version 3.8, but if you need me to I can look in to upgrading it.

eranif:
No need to upgrade it, on my Debian box, I have `clang-format-3.5` . So I was aiming for the the lowest denominator `.clang-format` file

PR #1698, opened 2017-07-20
fixed "error: ‘for_each’ is not a member of ‘std’"

SteelPh0enix:
Fixed "error: ‘for_each’ is not a member of ‘std’", preventing CodeLite to compile


eranif:
Duplicate of https://github.com/eranif/codelite/pull/1692

Issue #1710, opened 2017-08-07
include <wx/wx.h> before <wx/chartype.h> in codelite_vim/vimCommands.h


zufuliu:
compiler error SIZEOF_WCHAR_T must be defined before including this file in wx/defs.h

zufuliu:
build CodeLite (latest git) against wxWidgets (latest git) on Windows produce above error.
see wx/chartype.h line 178 at https://github.com/wxWidgets/wxWidgets/blob/master/include/wx/chartype.h#L178
which says:
```c
#ifndef SIZEOF_WCHAR_T
    #error "SIZEOF_WCHAR_T must be defined before including this file in wx/defs.h"
#endif
```
for windows `SIZEOF_WCHAR_T` is defined at wx/defs.h line 896:
https://github.com/wxWidgets/wxWidgets/blob/master/include/wx/defs.h#L896
```c
        #ifndef SIZEOF_WCHAR_T
            /* Windows uses UTF-16 */
            #define SIZEOF_WCHAR_T 2
        #endif
```

`gcc -v`:
```
Thread model: posix
gcc version 7.1.0 (x86_64-posix-seh-rev1, Built by MinGW-W64 project)
```

```
$ gcc -dM -E - < /dev/null | grep WCHAR
#define __WCHAR_MAX__ 0xffff
#define __WCHAR_MIN__ 0
#define __WCHAR_WIDTH__ 16
#define __GCC_ATOMIC_WCHAR_T_LOCK_FREE 2
#define __WCHAR_TYPE__ short unsigned int
#define __SIZEOF_WCHAR_T__ 2
```

zufuliu:
this can be close now, however I get some warnings in vimCommand:
```
F:/Libs/codelite/codelite_vim/vimCommands.cpp: In member function 'bool VimCommand::OnNewKeyDown(wxChar, int)':
F:/Libs/codelite/codelite_vim/vimCommands.cpp:49:11: warning: enumeration value 'SEARCH_CURR_MODUS' not handled in switch [-Wswitch]
     switch(m_currentModus) {
           ^
F:/Libs/codelite/codelite_vim/vimCommands.cpp:49:11: warning: enumeration value 'ISSUE_CMD' not handled in switch [-Wswitch]

F:/Libs/codelite/codelite_vim/vim_manager.cpp: In member function 'void VimManager::OnKeyDown(wxKeyEvent&)':
[...]

PR #1716, opened 2017-08-13
codeVim

StefanoBelloni:
Hi Eranif
* I implemented some new vi-commands, 
* there is a pop up a status bar for the ':'-commands 
* eliminate the wxchartype.h include: this should solve the  #issue https://github.com/eranif/codelite/issues/1710, although I never experienced such an error myself.
All the best
Stefano.


eranif:
Hi,
This can not be merged into CodeLite due to conflicts.
Can you please recreate the PR? (also, please exclude the `.gitignore` file from the PR)

Thanks!

eranif:
Hi @StefanoBelloni , I have tested the VIM plugin and it seems kinda buggy atm.
* Fixed a crash that occurs in the following scenario: 
** Have VIM plugin disabled
** Open a file in CodeLite
** Search something using the `/` command
** Hit ENTER -> crash
The crash was due to NULL `m_ctrl` member in `VimCommand`, this can happen because the `editor-changed` event is sent while the plugin was disabled so you ignored it and did not call `VimCommand::set_ctrl` function

* Buggy behavior:
** Open multiple files
** Enter NORMAL mode
** Close the first file so the focus is set to the next file
** In the second file, the mode seems to be INSERT (no block cursor)

* Unable to delete lines using `dd` command
* I also got it to hang, but I am not sure which key combinations I hit and I was not running under the debugger so I could not trace down that issue

PR #1752, opened 2017-09-11
Fix some missing includes in cProjectDependecySorter

frankivo:


dghart:
Compilation fix for gcc 7.2.0

PR #1815, opened 2017-11-12
[Debugger] Add stepi for GDB.

yaakuro:
Hi

I am not sure how to deal with the shortcuts or the icons. Feel free to adapt or leave them out. I hope you consider this PR. If there are things wrong or needs to be changed please feel free to ask.

Yaakuro


eranif:
This PR looks OK.
One minor thing:
Can you please remove the toolbar buttons?
The toolbar as it is now, already has too many buttons. The debugger contains the common actions (Next, Step, Stop, Start). The same was that we don't have "Next Instruction" in the toolbar - we should not add "Stepi" into the toolabr

Thanks

yaakuro:
OK, finished.

PR #1837, opened 2017-11-21
Fix typo in QMake settings screen.

JeroenSack:


JeroenSack:
I hope I generated the .mo files the right way, I see quite a lot of changed in the binary format, so might be good if someone could check that. I used "msgfmt -o <.mo file> <.po file>".

PR #1850, opened 2017-12-11
Vim Plugin

StefanoBelloni:
Hallo,
I made a small improvement in VimPlugin
- fix bug: Enter in NORMAL MODE divide the current line.
- Add information when searching in text.

All the Best
Stefano.


codelite_vim/vimCommands.cpp
```diff
@@ -8,7 +8,7 @@
 /*EXPERIMENTAL*/
 #include "wx/uiaction.h"
 
-VimCommand::VimCommand()
+VimCommand::VimCommand(IManager* m_mgr)
```
eranif: Just an FYI:
You can call `clGetManager()` (defined in `globals.h`) to get the instance of the `IManager` without passing it around

PR #1855, opened 2017-12-17
Added algorithm inclusion

aviallon:
Added algorithm inclusion since it's required for std::for_each.
(I had compile error without this inclusion)


eranif:
Why are you fixing the `optimized_cxx_project` branch? If any, you should build the `master` branch
`optimized_cxx_project` was already merged into `master` and is obsolete

PR #1909, opened 2018-02-17
Added a C++17 option

aviallon:
Added a C++17 option for code completion and compile options, since it's now completely supported by main compilers like GCC and VCC.


eranif:
This branch can not be merged due to conflicts.
This is probably because some of the changes were already submitted with an earlier PR

aviallon:
Oops that's true. Forgot to pull upstream changes :D. I'll fix that right know.

PR #1911, opened 2018-02-20
Sort tabs alphabetically in clTabCtrl drop down?

lukester1975:
Hi

I find I only use the tab view drop down when I've got lots of tabs open (i.e. more than can fit across the screen), but then can spend a long time looking up and down the list for the file I want since it's not sorted. Obviously the tabs tab in the workspace view pane can be sorted, but I personally never use that.

Anyway, just a suggestion and here's a go at implementing it...

Thanks

### DIFF
--- a/Plugin/Notebook.cpp
+++ b/Plugin/Notebook.cpp
@@ -812,7 +812,7 @@ void clTabCtrl::OnMouseMotion(wxMouseEvent& event)
         wxString pagetip = m_tabs.at(realPos)->GetTooltip();
         if(pagetip != curtip) { SetToolTip(pagetip); }
     }
-    
+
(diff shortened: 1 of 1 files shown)


aviallon:
That's a cool feature. I personally like it since I always have tons of tabs opened x)

dghart:
Hi,
Though I prefer to group my tabs, I can see why this change will often be desirable. However it can also be an anti-feature: e.g. CodeLite has multiple CMakeList.txt files in different projects. I currently have only 2 of those loaded, but I've often had 3 or 4. With alphabetical sorting there's no way to distinguish between them.

So in 6b13e6d I've made this optional. It defaults to the original, non-alphasorted, behaviour; you can change it in Settings > Preferences > Windows & Tabs.

PR #1932, opened 2018-03-05
spellcheck: reduce continuous spell checking

lukester1975:
Hi 

The continuous spell check option can burn CPU time when it doesn't really need to, and then I made it worse with #1928 :blush: which I see you've merged (thanks!).

So I've had a go at reducing that CPU usage. There are two versions here: the first one simply doesn't run the check if the editor doesn't report IsModified() == true since the last time it ran. Very simple but obviously still wastes CPU time if the file stays unsaved but not further modified.

The second commit hooks in to wxEVT_STC_MODIFIED to store the time an actual modification was made, and only runs the spell checker if it hasn't run since the last modification.

I'm no GUI programmer let alone wxwidgets so I'm not sure if that's the best/only option; maybe there's a better way?

Thanks!

Luke.

### DIFF
--- a/Interfaces/ieditor.h
+++ b/Interfaces/ieditor.h
@@ -96,6 +96,11 @@ public:
     virtual bool IsModified() = 0;
 
     /**
+	 * @brief Get the editor's modification count
+	 */
+    virtual wxUint64 GetModificationCount() const = 0;
+
+    /**
      * \brief return the current editor content
      */
     virtual wxString GetEditorText() = 0;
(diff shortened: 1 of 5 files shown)


eranif:
Class `LEditor` already handles `wxEVT_STC_MODIFIED` event.
It is called `OnChange` (see: https://github.com/eranif/codelite/blob/master/LiteEditor/cl_editor.cpp#L4175 )

Please use this handler instead of binding another one, thanks!

lukester1975:
Ah cool, missed that - thanks!

Here's an updated version, plus just using a counter rather than time to track modifications.

PR #1939, opened 2018-03-11
Optional copying entire line with empty selection?

lukester1975:
Hello again

Being fat-fingered, I'm often hitting ctrl-c when I mean ctrl-v. That is unfortunate as it replaces the clipboard contents with the current line (well, if not pasting in to a selection).

I've generally just hacked CopyAllowLine() -> Copy() in EditHandler::ProcessCommandEvent() but thought I'd tidy it up and make it a proper option like with #1927.

Sorry the menu_event_handlers.cpp diff is the whole file - I guess line endings have been normalized or something. It's just:

```
    if(event.GetId() == wxID_COPY) {
        if(options->GetCopyLineEmptySelection())
            editor->CopyAllowLine();
        else
            editor->Copy();
```

instead of always CopyAllowLine().

Thanks!

Luke.


eranif:
I am OK with this feature. But placing it under `Caret & Scrolling` in the UI does not seems really a good place. `Ctrl-C` / `Ctrl-V` behaviour should be located under the `Edit` page. Can you please move the UI part to this page instead?

lukester1975:
Sure can do, like this? 

image

eranif:
Yes, like this

lukester1975:
Cool. I've updated my branch.

Thanks
Luke.

Issue #1668, opened 2017-06-13
Old keyboard shortcut would not be replaced


BenQoo:
The reproduce steps as follow:

1. As the default setting:
`[Help] - [About]` without shortcut assigned.
`[Edit] - [Delete Line]` assign `[Ctrl-L]`.

2. Set `[Help] - [About]` to `[Ctrl-L]`.

3. CodeLite will popup this window, then choose `Yes` to replace.
![codelite_shortcut](https://user-images.githubusercontent.com/29389982/27062401-5201827e-501d-11e7-82b6-f13c12997fc2.png)

4. `[Edit] - [Delete Line]` shortcut is still exist, not be cleared. (problem)
![codelite_replace](https://user-images.githubusercontent.com/29389982/27062483-e5534ef4-501d-11e7-838e-dd22b23a2376.png)

AJenbo:
I have the same issue, looks like some are correctly removed while others aren't.

PR #1942, opened 2018-03-13
Fix re-assigning accellerator when old accelerator is filtered out from list control

lukester1975:
Hi

Just noticed that when re-assigning an accelerator, the old one will only be cleared if it is visible in the list control (i.e. not if it's filtered out).

Thanks!

Luke.


lukester1975:
BTW was also reported in #1668

Issue #1904, opened 2018-02-14
LLDB session using konsole suffers from the "/usr/bin/sleep" crashed thing


lukester1975:
Hi

Using konsole as the terminal with an LLDB session suffers from the issue that clDebuggerTerminalPOSIX::Clear() works around, but LLDBPlugin doesn't use it.

So here's a quick hack to do so, which seems to do the trick for me. I'm guessing it might not be doing that right now though because of non-Posix (judging by the name of the class...)? Anyway in case it's any use:

[0001-LUKE-made-LLDBPlugin-use-clDebuggerTerminal-so-konso.patch.txt](https://github.com/eranif/codelite/files/1725217/0001-LUKE-made-LLDBPlugin-use-clDebuggerTerminal-so-konso.patch.txt)

Thanks

PS another way I worked around before this one was to make sleep always return 0. It's pretty ugly due to the command to watch for being different than the one that is executed, but attached that too if it's of any 
interest:

[0001-LUKE-hacked-sleep-to-never-return-non-0-so-konsole-d.patch.txt](https://github.com/eranif/codelite/files/1725216/0001-LUKE-hacked-sleep-to-never-return-non-0-so-konsole-d.patch.txt)

Issue #1895, opened 2018-02-08
Ending lldb session doesn't kill process


lukester1975:
Hello again

Another issue I've been having is that ending an LLDB session wouldn't kill the debugged process. I *think* this is because LLDBConnector::m_attachedToProcess is potentially unitialized, so most of the time it's non-0 and the shutdown procedure just detaches rather than kills.

I was still seeing this happen every now and then, so also put a sleep in before the codelite-lldb process is killed (and removed the -9). Not really sure it's the right solution but it seems to have done the trick for me. Attached in case it's of use!

Regards

Luke.

[0001-LUKE-attempt-to-ensure-lldb-kills-process-on-shutdow.patch.txt](https://github.com/eranif/codelite/files/1707750/0001-LUKE-attempt-to-ensure-lldb-kills-process-on-shutdow.patch.txt)

Issue #1894, opened 2018-02-08
lldb fails on first run


lukester1975:
Hi

I've been running in to an issue where an LLDB debugging session fails on first attempt. I've traced it down to the fact that LLDBConnector::m_debugServer isn't initialzed at the point LLDBPlugin::OnDebugStart() calls LLDBConnector::LaunchLocalDebugServer(). Subsequent debug sessions are OK as m_debugServer is now set.

I've attached my workaround, no idea if it's the correct solution really but it sorts the problem for me!

Note there were a couple of side effects of the issue I noticed:

1. Any custom perspective layout would be lost, since the shutdown code would save the perspective but it wouldn't actually have been loaded yet, so basically resetting to default.

2. Similarly with breakpoints - any set breakpoints would be lost as the LLDB shutdown would save them but they hadn't actually been set yet.

I guess with a fix for the startup issue these aren't really a problem.

Thanks!

Luke.

[0001-LUKE-fixed-issue-on-startup-where-LLDBPlugin-m_debug.patch.txt](https://github.com/eranif/codelite/files/1707725/0001-LUKE-fixed-issue-on-startup-where-LLDBPlugin-m_debug.patch.txt)

PR #1955, opened 2018-03-19
RFC: lldb fixes

lukester1975:
Hi

I've pulled together various lldb mods I've been running (some tidied up from #1894 #1895 #1904).

Seemed like individual PRs was a bit much but I can split it up if that's easier as there are quite a few commits. I think some feedback is probably necessary hence just asking for comments at this point really...

- lukester1975/codelite@777c3fe: nothing controversial there, just tidied up version of #1894.
- lukester1975/codelite@fbf200c: I mentioned this in #1894; lldb perspective getting messed up if the debug session doesn't start successfully. Not sure if this is the preferred method? Checking LLDBConnector::IsRunning() didn't seem quite right in case something else fails before that becomes true, but after the perspective is loaded. Perhaps a more generic solution would be to add PerspectiveManager::GetActivePerspective() or similar, then only save the perspective on shutdown if the active one is lldb?
- lukester1975/codelite@9735ba4: the simple bit from #1895 (missing initializer for m_attachedToProcess). I still run the other bit from #1895 (sleep before kill) but would like to figure out a nicer solution!
- lukester1975/codelite@9ae992a: maybe I should've split this up. There's two parts:

1. first is to search for x-terminal-emulator rather than just look in /usr/bin. That makes it the same as codelite itself, plus means I can put a symlink to konsole in my path called x-terminal-emulator :smile:
2. then I realised that wasn't really sufficient as konsole doesn't use -T to set the window title. So that's the reason for the additional check for a new codelite-debug-terminal-emulator, which I have in my path with contents:

```
#!/bin/bash
TITLE=$1

# Need the --separate so Codelite kills the terminal on debug end.
echo -n konsole --separate --hide-tabbar -p tabtitle=\"${TITLE}\"
```
Then finally due to fun with quoting, the actual invocation of the terminal command is wrapped with eval.

konsole is nice in that it ignores -T, but other terminals (e.g. gnome-terminal) aren't so nice. So perhaps this could do with some more work...?

- lukester1975/codelite@545d220: as in #1904 
- lukester1975/codelite@24de34e: I use kwin rules to keep the debugger console out of the way of Codelite itself. That was easy with gdb given the "Debugging: " prefix on window titles, but not with lldb. So this just normalized the titles across gdb/lldb. The only bit I wasn't sure about is is clDebuggerTerminal.cpp the right place for the title formatting functions?
- lukester1975/codelite@8241fba: just a spelling fix
- lukester1975/codelite@e432afc: I noticed I couldn't remove a watch I'd added if it matched a local. Easy fix, just took a while to track down.
- lukester1975/codelite@ffc052d: also the context menu in the locals display wasn't working.
- lukester1975/codelite@f2ee402: this one probably needs work. Whilst hacking around, I'd often manage to kill codelite-lldb which would cause the perspective to get confused (the lldb panels remain but in the default perspective). This was just a quick hack to do like if lldb itself dies, and inform the user and do a bit more shutdown work. I was thinking it should be a different event in order to show a better dialog (or some text in the event to pass to the dialog to explain what's happened?)

Anyway, hope that's all useful! Happy to make some more mods as you see fit. With all this though, I'm happily using lldb to debug. I've given up with gdb mostly - it's just so slow (gdb itself, not the codelite integration).

Thanks

Luke.


LLDBDebugger/LLDBPlugin.h
```diff
@@ -33,6 +33,7 @@
 #include "LLDBProtocol/LLDBConnector.h"
 #include "LLDBProtocol/LLDBEvent.h"
 #include "LLDBProtocol/LLDBRemoteConnectReturnObject.h"
+#include "../Plugin/clDebuggerTerminal.h"
```
eranif: I think we can simply include `#include <clDebuggerTerminal.h>` there is no need for the relative path here

LLDBDebugger/LLDBProtocol/LLDBConnector.cpp
```diff
@@ -438,7 +440,9 @@ void LLDBConnector::OnProcessOutput(clProcessEvent& event)
 void LLDBConnector::OnProcessTerminated(clProcessEvent& event)
 {
     wxDELETE(m_process);
-    Cleanup();
+
+    clDebugEvent e2(wxEVT_LLDB_CRASHED);
+    ProcessEvent(e2);
```
eranif: Are we sure that `wxEVT_LLDB_CRASHED` is the correct one to send here? This will trigger a message to the user saying that LLDB has crashed. We also have `wxEVT_LLDB_EXITED` (`OnProcessTerminated` can be called also if the user kills LLDB and not crashed)

Also, I would prefer to fire the event using `AddPendingEvent` and not `ProcessEvent` (the `AddPendingEvent` schedules the event in an async mode unlike `ProcessEvent` which processes the event in sync mode)

LLDBDebugger/LLDBPlugin.h
lukester1975: Cool. Not sure why I did that!

LLDBDebugger/LLDBProtocol/LLDBConnector.cpp
lukester1975: Yeah like I say I did wonder if it was the wrong event, but also thought a dialog box popping up would be useful info so I just hacked it to use wxEVT_LLDB_CRASHED. I can add another event, or I guess some user data can be attached to the event (e.g. "lldb" or "codelite-lldb", as appropriate) that the dialog box can use?

Sync/async - OK. The idiosyncrasies of wx events aren't clear to me yet :smile: I just did sync so it was effectively the same as the existing call to Cleanup() (as I understood it).

Cheers!

eranif: You can use `wxEVT_LLDB_EXITED` (this sounds about right to me)

lukester1975: So no message box you mean, since LLDBPlugin already binds wxEVT_LLDB_EXITED?

lukester1975:
Oh! I hadn't done any of the changes mentioned yet, shall I do another PR with them? Probably wouldn't be until the weekend though.

eranif:
its ok.  i applied the PR so i could test it locally.  no rush :)

lukester1975:
OK found a few minutes... it failed for me too, but then I realised I had no terminal installed! It was fine once I installed xterm. 

I assume something else was going on in your case?

If I add

```
exec &>>~/out.txt
set -xe
```
to the top of codelite_xterm I get something like:

```
+ program_title='Debugging: "/path/to/exe" '
+ which codelite-debug-terminal-emulator
++ which x-terminal-emulator
+ terminal=/usr/bin/x-terminal-emulator
+ '[' 0 -ne 0 ']'
+ export LD_LIBRARY_PATH
+ eval /usr/bin/x-terminal-emulator '' -T '"Debugging:' '"/path/to/exe"' '"' -e '/bin/sleep 86325'
```

lukester1975:
Ah just seen #1972 , I'll look at the LD_LIBRARY_PATH case.

PR #1959, opened 2018-03-24
lldb-fixes follow up

lukester1975:
Hi

I did a bit more looking in to the notification of codelite-lldb crashing thing. I noticed LLDBNetworkListenerThread had code to cope with losing the connection to codelite-lldb but it wasn't triggering - it looks like clSocketBase::ReadMessage() wasn't coping with 0 length recv()s for the header or body.

I also made a change here to have a separate clSocketBase::GetLastError() function and error(const int errorCode) to avoid having to worry about whether errno gets reset between reading it and wanting to convert it to a string (since the code now has a check for EWOULDBLOCK before calling error()). Not really an issue here since strerror() is presumably compliant on all platforms Codelite will run on, it's just my personal preference. Happy to back it out if you prefer :smile:

I thought fixing the socket error notification might mean the previous mod to send wxEVT_LLDB_CRASHED from LLDBConnector::OnProcessTerminated() wouldn't be needed, but it is due to racing between the socket notification and process termination. I think what's happening is if the process termination is received first and it just calls Cleanup(), that could terminate the LLDBNetworkListenerThread such that it never generates wxEVT_LLDB_CRASHED and then the debugger just closes with no message box. If the socket notification is received first, it was OK.

So I've left both sources of wxEVT_LLDB_CRASHED in (socket close and process termination) and made OnLLDBCrashed() a bit nicer with re-entracy (to avoid multiple message boxes). I also noticed LLDBConnector::m_goingDown was never reset so added that.

The include tidy up is also in here.

Anyway, see what you think!

Cheers

Luke.

### DIFF
--- a/CodeLite/SocketAPI/clSocketBase.cpp
+++ b/CodeLite/SocketAPI/clSocketBase.cpp
@@ -27,6 +27,7 @@
 #include <cerrno>
 #include <cstdio>
 #include <sstream>
+#include <memory>
 
 #ifndef _WIN32
 #include <fcntl.h>
@@ -54,7 +55,7 @@ void clSocketBase::Initialize()
 }
 
 // Read API
-int clSocketBase::Read(wxMemoryBuffer& content, long timeout) 
+int clSocketBase::Read(wxMemoryBuffer& content, long timeout)
 {
     content.Clear();
 
@@ -67,8 +68,12 @@ int clSocketBase::Read(wxMemoryBuffer& content, long timeout)
             memset(buffer, 0x0, sizeof(buffer));
             int bytesRead = recv(m_socket, buffer, sizeof(buffer), 0);
             if(bytesRead < 0) {
-                // Error
-                throw clSocketException("Read failed: " + error());
+                const int err = GetLastError();
+
+                if(eWouldBlock != err) {
+                    // Error
+                    throw clSocketException("Read failed: " + error(err));
+                }
 
             } else if(bytesRead == 0) {
                 // connection closed
@@ -88,7 +93,7 @@ int clSocketBase::Read(wxMemoryBuffer& content, long timeout)
     return kTimeout;
 }
 
-int clSocketBase::Read(wxString& content, const wxMBConv& conv, long timeout) 
+int clSocketBase::Read(wxString& content, const wxMBConv& conv, long timeout)
 {
     wxMemoryBuffer mb;
     int rc = Read(mb, timeout);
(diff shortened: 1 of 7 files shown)


CodeLite/SocketAPI/clSocketBase.cpp
```diff
@@ -239,36 +271,28 @@ int clSocketBase::ReadMessage(wxString& message, int timeout)
     message_len = ::atoi(msglen);
 
     bytesRead = 0;
-    char* buff = new char[message_len + 1];
-    memset(buff, 0, message_len + 1);
+    std::unique_ptr<char> pBuff(new char[message_len]);
```
lukester1975: I couldn't see the delete of buff in the success case so switched to a scoped array.

CodeLite/SocketAPI/clSocketBase.cpp
```diff
@@ -239,36 +271,28 @@ int clSocketBase::ReadMessage(wxString& message, int timeout)
...
     int bytesLeft = message_len;
     int totalRead = 0;
     while(bytesLeft > 0) {
-        rc = Read(buff + totalRead, bytesLeft, bytesRead, timeout);
+        rc = Read(pBuff.get() + totalRead, bytesLeft, bytesRead, timeout);
         if(rc != kSuccess) {
-            wxDELETEA(buff);
             return rc;
 
-        } else if(rc == 0) {
```
lukester1975: Should've been a check on bytesRead == 0 presumably but instead I moved the check and throw to Read(), so the Read() used for the message_len also copes with socket being closed.

LLDBDebugger/LLDBProtocol/LLDBConnector.h
```diff
@@ -70,10 +72,10 @@ class LLDBConnector : public wxEvtHandler
 
 protected:
     clSocketClient::Ptr_t m_socket;
-    LLDBNetworkListenerThread* m_thread;
+    std::unique_ptr<LLDBNetworkListenerThread> m_thread;
```
lukester1975: I made this a unique_ptr so Cleanup() was safe(r) to call multiple times, but then made the mod in LLDBPlugin::OnLLDBCrashed() to SetGoingDown() so it's not really needed I guess. But seems nicer :smile:

LiteEditor/mainbook.cpp
```diff
@@ -612,11 +612,12 @@ LEditor* MainBook::OpenFile(const wxString& file_name, const wxString& projectNa
             AddPage(editor, fileName.GetFullName(), tooltip.IsEmpty() ? fileName.GetFullPath() : tooltip, bmp);
         }
         editor->SetSyntaxHighlight();
+        ManagerST::Get()->GetBreakpointsMgr()->RefreshBreakpointsForEditor(editor);
```
lukester1975: Dunno if this is the right thing to do or the right place to do it, but seems to work!

PR #1969, opened 2018-04-03
RFC: accelerators on floating pane

lukester1975:
Hi

I noticed whilst using LLDB with a floating locals/watch pane that the keyboard accelerators weren't working when the floating pane is active. A quick google suggested SetAcceleratorTable() on the pane's frame would workaround it, but then I discovered clMainFrame::OnAuiManagerRender() does just that - but is disabled (since 67f6972babd19fb44853f9aed928024932f66e59 8 years ago).

I commented it back in and it appeared to work but no doubt there's a reason why it's disabled! As I've said before I'm a complete wxWidgets newb so I've no idea what that reason would be :smile:

EVT_AUI_RENDER does seem to fire a lot (hardly surprising I suppose!), so I had a hack around for an alternative implementation which is this branch. The idea being to only copy the table when the window is activated and floating. But maybe there are windows which this shouldn't be done to? Is code specific to the LLDB panes better? The hack to get the main frame is also unpleasant; I couldn't see an obvious way to get at it from clDockingManager without coupling it more (e.g. passing in as a wxWindow to the constructor).

If you think this is viable or needs a few changes I'm more than happy to make them!

Thanks

Luke.


Plugin/clDockingManager.cpp
```diff
@@ -42,6 +45,32 @@ void clDockingManager::OnButtonClicked(wxAuiManagerEvent& event)
...
+        // HACK find the top level frame.
+        // TODO or pass clMainFrame (as a wxWindow) to the ctor?
+        wxWindow* parent = pane->frame;
+        auto grandParent = parent;
+
+        while(grandParent) {
+            parent = grandParent;
+            grandParent = parent->GetParent();
+        }
+
```
eranif: You could simply use `wxGetTopLevelParent`

eranif:
IIRC, `wxTheApp->GetTopWindow()` will always return the main frame (`clMainFrame`) this is because we explicitly set it in `app.cpp`.
We could restore the automatic behavior (find the first top window from the current window list) by calling `wxApp::SetTopWindow(NULL)`

see here: http://docs.wxwidgets.org/3.1.1/classwx_app.html#a39033ec4c79bc1871c91ada7b31941e6
Also, we don't need to pass `IManager` you can access it by calling `::clGetManager()`
But feel free to choose whatever works :)

PR #1973, opened 2018-04-08
lldb: added some features

lukester1975:
Hello Eran

Here's a few features for lldb that can hopefully be integrated. A few notes:

- lukester1975/codelite@0331d6d: I use prctl(PR_SET_NAME, gettid(), "nice name") to set per-thread names, so it's nice (essential really) to see them in the debugger.
- lukester1975/codelite@3e25771: If you've expand a variable quite deep to get to some "interesting" fields, it's a shame to lose that expansion after stepping. So this just adds expansion beyond top-level (and handles collapsing).
- lukester1975/codelite@b2a5a0a: This might be controversial so I can remove it from the PR if you prefer. It's just my personal opinion that the columns should have the same order in each pane :smile:

Oh I guess the only other controversial thing might be the use of anonymous namespaces for a few file-scope things, since I read in the style guide you don't use namespaces. Anonymous ones are a slightly different case though I imagine...

Let me know if there's anything that needs fixing up to make it good to merge!

Thanks

Luke.

### DIFF
--- a/LLDBDebugger/LLDBCallStack.cpp
+++ b/LLDBDebugger/LLDBCallStack.cpp
@@ -26,14 +26,16 @@
 #include "LLDBCallStack.h"
 #include "LLDBProtocol/LLDBEvent.h"
 #include "LLDBProtocol/LLDBConnector.h"
+#include "LLDBPlugin.h"
 #include <wx/wupdlock.h>
 #include "macros.h"
 #include "globals.h"
 #include "file_logger.h"
 
-LLDBCallStackPane::LLDBCallStackPane(wxWindow* parent, LLDBConnector* connector)
+LLDBCallStackPane::LLDBCallStackPane(wxWindow* parent, LLDBPlugin& plugin)
     : LLDBCallStackBase(parent)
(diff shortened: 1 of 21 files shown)


lukester1975:
Added another commit to this branch though it's a fix rather than a feature - I tripped over stepping not working properly with multiple threads. After stepping, the selected thread would always be reset back to the main thread. Was a head scratcher for a while!

So looks like after stepping the StopReason will be eStopReasonPlanComplete, so that needs coping with too when choosing which thread to select.

I made it check for the other StopReasons too and choose the "best" thread to select based on the stop reasons across all threads, though whether you can actually get >1 thread with an "interesting" StopReason at the same time I'm not sure. As such it needed to prioritize StopReasons so I just made something up (see StopReasonToPriority()).

Also, since I made it less efficient I took the opportunity to save having to iterate the thread list twice (or one potentially partial iteration + one full iteration as it was previously).

Thanks

Luke.

lukester1975:
Another little one to make the selected thread and frame visible when initial populated.

PR #1974, opened 2018-04-08
Web search context menu?

lukester1975:
Hi

What do you think about a context menu option to search for selected text with the default web browser? I find it incredibly helpful... especially when trying to get my head around wxWidgets and the lldb API :smile:

Would it be better in the help plugin, with the context menu option under the "Search the docs for..." item rather than "Search for selected text in workspace"?

I just added the ctrl-shift-w accelerator as it seemed unused by default.

Thanks

Luke.


eranif:
It often cross my mind :)
It is a great idea

eranif:
This PR could not be merged due to conflicts - can you please re-submit it?

lukester1975:
Sure, here you go.

PR #1988, opened 2018-04-24
Spellcheck context menu tweak

lukester1975:
Hi

I realised after making the continuous spell check option more persistent that if a word is misspelled, the editor context menu is completely replaced with the spell check context "suggestions" menu when right-clicking on that word - which is a bit unhelpful if you want to leave the word as-is but do something else from the regular context menu.

So here's a mod to make the suggestions appear in the regular spell-check content menu. I had to add UnHookPopupMenu() support (6c22d66) in order to be able to re-start the spell-check timer.

If there are spelling suggestions then the spell-check sub-menu gets put at the top of the context menu (or the top until some other plugin puts something else on top of it...). I don't know if that's controversial or not :smile:

Also, because of using HookPopupMenu() rather than hooking wxEVT_CONTEXT_MENU_EDITOR, the spell-check sub-menu will appear in a slightly different place when there are no suggestions. I guess it could use wxEVT_CONTEXT_MENU_EDITOR and not HookPopupMenu() but leave UnHookPopupMenu() in order to reset the timer if that's a concern. It just didn't seem very symmetrical that way! 

Anyway, if you think it's a good idea and there are any tweaks needed let me know...

Thanks

Luke.

### DIFF
--- a/CallGraph/callgraph.cpp
+++ b/CallGraph/callgraph.cpp
@@ -205,25 +205,6 @@ void CallGraph::HookPopupMenu(wxMenu* menu, MenuType type)
 
 //-----------------------------------------------------------------------------
 
-void CallGraph::UnHookPopupMenu(wxMenu* menu, MenuType type)
-{
-    if(type == MenuTypeEditor) {
-        // TODO::Unhook items for the editor context menu
-    } else if(type == MenuTypeFileExplorer) {
-        // TODO::Unhook  items for the file explorer context menu
-    } else if(type == MenuTypeFileView_Workspace) {
-        // TODO::Unhook  items for the file view / workspace context menu
-    } else if(type == MenuTypeFileView_Project) {
-        // TODO::Unhook  items for the file view/Project context menu
-    } else if(type == MenuTypeFileView_Folder) {
-        // TODO::Unhook  items for the file view/Virtual folder context menu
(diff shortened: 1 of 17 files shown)


lukester1975:
Sure, here's a v2 attempt. Not sure if wxEVT_CONTEXT_MENU_EDITOR_END is a good name for the event I added after the popup closes?

Since UnHookPopupMenu() is not implemented, there's a commit here removing all of them (it confused me when doing v1 why it was never called :smile:). I can remove that if you prefer!

Thanks

Luke.

eranif:
Why do we need the 'END' event?
You can use the event `wxEVT_MENU_CLOSE` and bind it to `wxTheApp`

Something like this will do the trick (it will catch **any** menu closed, but we don't really care since its not a frequent action):

```C++
wxTheApp->Bind(wxEVT_MENU_CLOSE, [](wxMenuEvent& e) { clWARNING() << "Menu has been closed!"; }, wxID_ANY);
```

lukester1975:
No reason other than not knowing wxWidgets... Seems like wxEVT_MENU_CLOSE would fire before a menu selection which I guess is a bit different.

That said, looking at the code again I can't really see a reason for turning off the timer (especially with the check on "modification count" that's there now). So how about this v3 attempt? Just leaves the timer running...

PR #1993, opened 2018-04-28
lldb: locals/watches features

lukester1975:
Hello Eran

I've had a go at adding a few features to locals & watches in lldb:

- If you've expanded a variable down a lot of levels but are only really interested in the nested value, it's nice to be able to add a watch on that nested value then close down the top level variable easily. So I've added that as a content menu option, or you can drag a value out in to some empty space (just copying what VS does there and I was interested in seeing how to do such a thing with wxWidgets).
- Edit value support.
- Ability to set the display format for variables.

 See what you think but there are at least a few things that need some feedback if you have time:

- The clEditTextCtrl::Destroy() mod - I'm sure this won't be quite right (especially given the comment in there about >=2.9 wxWidgets and ScheduleForDestroy()). Without it though, you can edit a value once with F2 but then not again (though clicking to edit works multiple times). Also all items backgrounds stay grey rather than blue, like they are in edit mode.

- Speaking of editing with F2 - this only works if I remove F2 from the global accelerators table. Is there a way to override for a specific control so if the locals view is active, F2 will edit an item otherwise the global accelerator is used?

- Adding a watch on an expanded item needs a fully qualified expression to watch. The existing "itempath" in LLDBLocalsView almost worked but not quite (always uses . between elements whereas pointer types will naturally need ->). So I added the result of SBValue::GetExpressionPath() to LLDBVariable, which can then be passed around. Strangely though the result of GetExpressionPath() is not always watchable - e.g. *argv will have GetExpressionPath() of "*(argv)" which fails GetValueForVariablePath()! Seems like a flaw in the LLDB API itself to me so I've not tried to do anything smarter...

- For the display variable format thing, I hacked the format into LLDBCommand using the existing frameId member. Not very pretty so would it be better to add a specific member?

I think that's it!

Luke.

### DIFF
--- a/LLDBDebugger/LLDBLocalsView.cpp
+++ b/LLDBDebugger/LLDBLocalsView.cpp
@@ -44,6 +44,29 @@ const int lldbLocalsViewEditValueMenuId = XRCID("lldb_locals_view_edit_value");
 const int lldbLocalsViewAddWatchContextMenuId = XRCID("lldb_locals_view_add_watch");
 const int lldbLocalsViewRemoveWatchContextMenuId = XRCID("lldb_locals_view_remove_watch");
 
+struct VariableDisplayFormat
+{
+    lldb::Format    m_format;
+    const char *    m_name;
+};
+
+constexpr VariableDisplayFormat lldbVariableDisplayFormats[] = {
+    {   lldb::eFormatDefault,           "Default"           },
+    {   lldb::eFormatDecimal,           "Decimal"           },
+    {   lldb::eFormatHex,               "Hex"               },
+    {   lldb::eFormatOctal,             "Octal"             },
+    {   lldb::eFormatBinary,            "Binary"            },
+    {   lldb::eFormatFloat,             "Float"             },
+    {   lldb::eFormatComplex,           "Complex"           },
+    {   lldb::eFormatBoolean,           "Boolean"           },
+    {   lldb::eFormatBytes,             "Bytes"             },
+    {   lldb::eFormatBytesWithASCII,    "Bytes (with text)" },
+    {   lldb::eFormatCString,           "C-string"          },
+    {   lldb::eFormatPointer,           "Pointer"           }
+};
+constexpr size_t lldbNumberOfVariableDisplayFormats = sizeof(lldbVariableDisplayFormats) / sizeof(lldbVariableDisplayFormats[0]);
+const int lldbLocalsViewVariableDisplayFormatId = wxWindow::NewControlId(lldbNumberOfVariableDisplayFormats);
+
 }
 
 LLDBLocalsView::LLDBLocalsView(wxWindow* parent, LLDBPlugin* plugin)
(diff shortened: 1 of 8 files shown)


lukester1975:
Very brave again :smile: Just wanted to check you saw my note about clEditTextCtrl::Destroy() above - I've no idea if it's the right thing to do...

eranif:
FWIW, it does not compile on Windows / Linux with older lldb version. I am attempting to fix this now :)

lukester1975:
Ah right. I've never managed to get a Windows build environment set up for Codelite I'm afraid! Maybe things clashing with cygwin but I will have another go one day.

Should be easy for me to set up a VM with a Linux distro with an older LLDB though - what is the oldest version you'd like to support and I can make sure any future mods compile on it before sending a PR?

eranif:
Some points to take care:
1. The current design is that **only** CodeLiteLLDBApp has access to the various `lldb` core API (e.g. `<lldb/SBBreakpoint.h>`)
2. CodeLite executable does not link against `liblldb.so` (it does not exist on all platforms)
3. the LLDBProtocol serialises the request in a classes known to CodeLite - it's up for CodeLiteLLDBApp to convert it into the proper lldb calls.

I am about to push a commit that addresses these issues

lukester1975:
Ah OK, I did mean to test it built with -DENABLE_LLDB=0 but completely forgot. Sorry about that.

eranif:
OK, this is now fixed.
I have refactored this feature. You can see my changes in this commit:

https://github.com/eranif/codelite/commit/fbee37dc3a27273ac1aa1f4b7164b96e4c894f96

eranif:
It seems that we have differences in the formatter settings, so many whitespaces changes entered the diff making it hard to read :/

lukester1975:
Ah OK so it is for building without the LLDB API being available at all, but to still be able to connect to a remote lldb-server - makes sense.

Yeah I've been avoiding using the .clang-format in the repo because of the number of changes it made (to keep the diffs small), but trying to stick the style on the wiki (probably badly, the Codelite style is quite different from my own!).

What do you think about changing AllowShortIfStatementsOnASingleLine to false? The reasoning being you can't set a breakpoint on the conditional statement as things are :disappointed: 

Also similarly AllowShortFunctionsOnASingleLine -> Empty and AllowShortBlocksOnASingleLine -> false?

Also one last one - long line breaking style. I've seen both align to open ( and also additional indent so wasn't sure which to use. E.g.

```
    wxTheApp->Disconnect(
        wxID_SELECTALL, wxEVT_UPDATE_UI, wxUpdateUIEventHandler(OutputTabWindow::OnEditUI), NULL, this);

    m_tb->AddTool(XRCID("hold_pane_open"),
                  _("Keep open"),
                  bmpLoader->LoadBitmap("ToolPin"),
                  _("Don't close this pane when an editor gets focus"),
                  wxITEM_CHECK);
```

What do you think about AlignAfterOpenBracket: DontAlign (like the top example)? That way if the opening ( is near the end of the line you don't end up with lots of wasted space.

Like the attached:  [clang-format.txt](https://github.com/eranif/codelite/files/1961930/clang-format.txt)

Regards

Luke.

PR #1995, opened 2018-05-01
Include <algorithm> to fix build

kekePower:
Hi.

Been unable to build for a few days and a google search lead me to this "fix" which helped me.


aviallon:
I confirm this bug. I just tried to compile it with default options on a clean machine and apparently <algorithm> include is missing

aviallon:
Thanks that's better :smile:

PR #2006, opened 2018-05-07
lldb: memory viewer.

lukester1975:
Hello Eran

Here's an attempt at a memory viewer for LLDB. It reuses the GDB memoryview.cpp for the UI, since everything in it GDB specific is in virtual functions that could be overridden (with one small exception).

I imagine things should be moved around a bit really? It has to

`#include "../LiteEditor/memoryview.h"`

which I guess is breaking the rules for plugins really! Better to move the common memory view UI code somewhere?

I also couldn't figure out how get a new pane to display if the perspective has a custom layout already, but I might investigate that further (I just deleted my lldb.layout to get the memory viewer to appear).

What do you think?

Regards

Luke.


lukester1975:
Just FYI since I noticed you doing some merging, I've got some updates to push to this branch I'll do later. Not that I think it's ready for merging anyway!

eranif:
I am currently fixing MemoryView by removing all direct access to `ManagerST`and all includes to `frame.h` and `manager.h` so I can move it to the `Plugin` library (which can be used by the all plugins etc)

What I did was:
I added two new events: `wxEVT_DEBUGGER_REFRESH_PANE` and `wxEVT_DEBUGGER_SET_MEMORY` so there is no need for direct calls to the debugger API

The next step will be to move the class `MemoryView` to the plugins directory (this will be done later today)

lukester1975:
Ah cool. I'll redo this on top of that when it's done then!

lukester1975:
Given the way you've done the new events, are you thinking of just reusing the existing memory view tab in the debugger notebook for LLDB? I have that kinda working though needs some tidying up!

Otherwise I guess it would be a bit weird potentially having two memory views open generating these events...

Thanks

eranif:
Yes, this was the idea. This class is now fully reuseable

PR #2026, opened 2018-05-30
Fixed a problem resulting in crashing when user update a new keyboard-shorcut.

snowxmas:


snowxmas:
codelite

eranif:
Thanks, I just added the actual fix without the rest of your commits

PR #2033, opened 2018-06-08
Removed all keyword "register" to adapt C++17 compiler.

snowxmas:


eranif:
**All** the files that you have fixed are auto generated.
Next time these files are generated, this fix will be lost...

Issue #2093, opened 2018-08-22
WxWidgets 3.0.x vs 3.1.x


kekePower:
Hi.

Please consider the usage of functions from WxWidgets 3.1.x since this is a development branch and I think that quite a few distros are conservative when they distribute packages.
In my case, Mageia will not distribute a development package unless it will become a stable package before the next major distro release and hence we're still using Wx 3.0.4.

I understand the need for and your desire to develop your application.

Cheers,
kekePower

eranif:
Because of this, we usually provide a pre-compiled binaries for wxWidgets (checkout our wiki page for more details)
I assume that you are facing an issue when building CodeLite by yourself, however, building wxWidgets is very easy on Linux (much easier than building CodeLite)
http://codelite.org/Developers/Linux

wx3.0.4 lacks some basic wxStyledTextCtrl features which are only available in wx3.1.X this is why we moved to wx3.1 and this can not be reverted

PR #2094, opened 2018-08-22
WxWidgets version check

kekePower:
Related to Issue #2093


eranif:
Maybe it allows the code to compile, but it breaks the functionality (i.e. instead of having a native choice, we will have an empty canvas)

PR #2193, opened 2018-12-15
clEditor: Support Relative Line Numbers

1123852253:


eranif:
I have merged this PR. In the future try to:

* rebase before creating PR (it avoids any mess in the repository history)
* some explanation about why this PR is needed. In this case, I simply assumed that it is useful for the VIM plugin (for example: easier to work with the DD / YY commands etc)
* I "enhanced" your code to show the current line in a more "visible" way
capture

1123852253:
Okay thank you
Adding this feature is really for better use of the VIM plugin

PR #2235, opened 2019-03-06
Added vector header dependent.

snowxmas:


hpmachining:
File is now in Codelite/LSP, not Codelite/json_rpc.

eranif:
This file no longer exists in CodeLite

Issue #2243, opened 2019-03-27
CMakePlugin add library path is not working


wevsty:
I found a problem when I compile with the CMake plugin provided in CodeLite.
I developed a library but added the library path in the Linker options and still can't find the library file.（Of course I am sure the library file exists in the right place.）
The error message is as follows
```
d:/mingw/bin/../lib/gcc/mingw32/8.2.0/../../../../mingw32/bin/ld.exe: cannot find -lmylib
collect2.exe: error: ld returned 1 exit status
```
CMakePlugin added the library path to CMAKE_LDFLAGS but this doesn't seem to work.
for example
```
# Library path
set(CMAKE_LDFLAGS "${CMAKE_LDFLAGS} -L. -L${WORKSPACE_PATH}/cmake-build-debug/output/ ")
```
I solved the problem after specifying the path using link_directories.
So i think using the link_directories command provided by CMake is a better solution.

Test environment：
OS: Windows 10
CMake 3.14.0
MinGW 8.2.0-3
CodeLite 12.0

wevsty:
Commit is merged.
Close issue.

Issue #2185, opened 2018-12-03
CMake: $(WorkspaceConfiguration) not translated to ${CONFIGURATION_NAME} in CMakeLists.txt


neildarlow:
Reproduced on codelite-12.0.10-gtk3 version.

When adding to the Linker Search Path in Project settings (either the Linker Search path setting or Global linker search path setting:
`$(WorkspacePath)/cmake-build-$(WorkspaceConfiguration)/output`

The $(WorkspaceConfiguration) is passed into CMakeLists.txt verbatim instead of being translated to ${CONFIGURATION_NAME} thus:
`# Library path`
`set(CMAKE_LDFLAGS "${CMAKE_LDFLAGS} -L. -L${WORKSPACE_PATH}/cmake-build-$(WorkspaceConfiguration)/output ")`

If you hack the Linker search path setting as follows:
`$(WorkspacePath)/cmake-build-${CONFIGURATION_NAME}/output`

you get the desired result of:
`# Library path`
`set(CMAKE_LDFLAGS "${CMAKE_LDFLAGS} -L. -L${WORKSPACE_PATH}/cmake-build-${CONFIGURATION_NAME}/output ")`

wevsty:
I have encountered a similar problem,I think the problem seems to happen in the step of custom build.

But i found the a commit https://github.com/eranif/codelite/commit/0de30a8bee3a097f6f42db59bf85f9cc2d24acb2 
I think this commit can solve this problem.

wevsty:
I found that I misunderstood the problem.
In order to fix this problem I added a PR.

eranif:
Thanks, fix merged

PR #2244, opened 2019-03-27
fix add libraries paths not working

wevsty:
fix Issue #2243. 
Change from modifying CMAKE_LDFLAGS to using link_directories().


wevsty:
update fix issue #2185.

PR #2302, opened 2019-07-14
Use PATH to search binaries instead of hardcoced /usr/bin/ path.

rozhuk-im:
At least FreeBSD by default installs all packages to /usr/local prefix.


rozhuk-im:
ping

PR #2345, opened 2019-09-28
Fixed some issues

1123852253:
FIX: GoTo Declaration error (When there are multiple identical keywords in the same file)
FIX: crash(Implement all Un-implemented Functions... )
LSP: add lua and rust LanguageId
LSP: Increased response speed (linux, stdio mode)

I have modified UnixProcess::ReadAll to solve the problem of lsp response speed.

### DIFF
--- a/CodeLite/UnixProcess.cpp
+++ b/CodeLite/UnixProcess.cpp
@@ -78,16 +78,27 @@ bool UnixProcess::ReadAll(int fd, std::string& content, int timeoutMilliseconds)
     int ms = timeoutMilliseconds % 1000;
 
     struct timeval tv = { seconds, ms * 1000 }; //  10 milliseconds timeout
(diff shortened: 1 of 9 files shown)


CodeLite/UnixProcess.cpp
```diff
@@ -78,16 +78,27 @@ bool UnixProcess::ReadAll(int fd, std::string& content, int timeoutMilliseconds)
...
+                continue;
+            }
+        } else if(rc == 0) {
+            // timeout
             return true;
         }
-    } else if(rc == 0) {
-        // timeout
-        return true;
+        break;
```
eranif: this is wrong. We should call here `return true` as it was before. timeout on select is OK.
Especially when you changed the code call `select` multiple times

PR #2359, opened 2019-10-16
Fix phpstan intergration

AJenbo:
The the `errorFormat` parameter name appears to have been changed.

Path are relative to CWD in phpstan https://github.com/phpstan/phpstan/issues/2019

Drop input for phpstan config as this wasn't fully implemented.

Assume that phpstan.neon exists for the current project root. This is
required because else the output will contain a notice that breaks the
XML. (note it's on stderr but we still include it in the output buffer for some reason)

### DIFF
--- a/PHPLint/phplint.cpp
+++ b/PHPLint/phplint.cpp
@@ -197,7 +197,8 @@ void PHPLint::QueuePhpstanCommand(const wxString& phpPath, const wxString& file)
     wxString phpstanPath = phpstan.GetFullPath();
     ::WrapWithQuotes(phpstanPath);
 
-    m_queue.push_back(phpPath + " " + phpstanPath + " analyze --errorFormat=checkstyle --no-progress " + file);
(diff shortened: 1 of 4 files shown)


PHPLint/phplint.cpp
```diff
@@ -279,6 +280,11 @@ void PHPLint::ProcessXML(const wxString& lintOutput)
 
     // Find the editor
     wxString filename = file->GetAttribute("name");
+    if(!filename.StartsWith("/")) {
```
AJenbo: This probably isn't working on Windows, but not sure how to tackle that?

eranif: Probably this is what you are looking for :

```c++
if(!wxFileName(filename).IsAbsolute()) {...}
```

AJenbo: Thanks, it's now integrated.

PR #2382, opened 2019-11-07
Use CMake to run the build process

AJenbo:
Using CMake to run the build should allow some more flexibility in what build tools are used.

Also store the resulting .deb file on the build server so that it can be downloaded and tested (nightly builds).
https://51-16066615-gh.circle-artifacts.com/0/root/repo/deb/CodeLite-13.0.5-gtk2---x86_64.deb

Note that `make -j` turned out to bleed RAM, using 32GB for about every ~25% of the process. Using nproc and physicalcpu lets the process use all cores but keep memory usage stable under 8GB (tested on 16 thread Linux system) for the whole process.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,8 +9,10 @@ jobs:
       - run: apt-get update -y
       - run: apt-get install -y libgtk2.0-dev pkg-config build-essential git cmake libssh-dev libwxbase3.0-dev libsqlite3-dev libwxsqlite3-3.0-dev
       - run: mkdir build-release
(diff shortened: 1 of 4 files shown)


eranif:
Looks like this can't be merged:

"All checks have failed"

AJenbo:
@eranif only for the last commit where i tried to make threads not hardcoded for the build server.

Note the check marks to the right:
image
I'll remove the last commit and re open it since that appears to be your only comment.

AJenbo:
@eranif please see, https://github.com/eranif/codelite/pull/2392

PR #2421, opened 2020-01-18
Gnome terminal does not work caused by 4e270fb

xukev:
Fix a regression. On Ubuntu 19_10. Gnome terminal does not work caused by 4e270fb. Change it back. 

Follow below steps to reproduce the regression:
1. Build and Make install codelite on Ubuntu 19_10.  Start this codelite and create a C++ console application.
2. Compile & Run.  Error happens when gnome starts up.


xukev:
@eranif Hi, please help review this change. Thanks.

PR #2696, opened 2021-03-22
Use correct help string for relative line preference option.

Jarod42:
Use correct help string for relative line preference option.


Jarod42:
Superseded by https://github.com/eranif/codelite/pull/2703/files

PR #2698, opened 2021-03-23
Fixed some issues 

liuanlin-mx:


git/GitDiffOutputParser.cpp
```diff
@@ -47,7 +54,7 @@ void GitDiffOutputParser::GetDiffMap(const wxString& rawDiff, wxStringMap_t& M,
 
                 // Reset the scanner loop
                 state = kLookingForFileName;
-                --i; // This will make sure we process this line again, however, in the other state switch case
+                nextToken = false; // This will make sure we process this line again, however, in the other state switch case
```
Jarod42, line 57: If that happens on last token, we don't loop again.

liuanlin-mx: thanks

PR #2709, opened 2021-04-17
Fix Output View multi-byte text handling

rlbxku1r:
1. Don't perform ToAscii() conversion
This will destroy all multi-byte character occurrences, but it's normal to see those texts in this Output Pane.
(e.g. source code comments, localized system messages etc.)
For this reason I've simply omitted the conversion.

2. Fix UTF-8 character width detection in the styling function

### DIFF
--- a/Plugin/clFindResultsStyler.cpp
+++ b/Plugin/clFindResultsStyler.cpp
@@ -136,9 +136,11 @@ void clFindResultsStyler::StyleText(wxStyledTextCtrl* ctrl, wxStyledTextEvent& e
     size_t matchStyleLen = 0;
     size_t i = 0;
     for(; iter != text.end(); ++iter) {
-        bool advance2Pos = false;
         const wxUniChar& ch = *iter;
-        if((long)ch >= 128) { advance2Pos = true; }
+        size_t chWidth = 1;
+        if(!ch.IsAscii()) {
+            chWidth = wxString(ch).mb_str(wxConvUTF8).length();
+        }
 
         switch(m_curstate) {
         default:
@@ -155,7 +157,7 @@ void clFindResultsStyler::StyleText(wxStyledTextCtrl* ctrl, wxStyledTextEvent& e
                 ctrl->SetStyling(1, LEX_FIF_DEFAULT);
             } else {
                 // File name
-                filenameStyleLen = 1;
+                filenameStyleLen = chWidth;
                 m_curstate = kFile;
             }
             break;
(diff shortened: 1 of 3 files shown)


Greg-21:
Since you're working on it, will this PR fix the issue shown below as well?

Broken-completion

rlbxku1r:
> Since you're working on it, will this PR fix the issue shown below as well?
> 
> Broken-completion

Unfortunately your problem seems unrelated with this PR.

eranif:
@Greg-21 can you post here a text snippet with the issue? (or better yet, open a separate bug report)

Greg-21:
Reported as #2710.

Plugin/clFindResultsStyler.cpp
```diff
@@ -155,12 +158,12 @@ void clFindResultsStyler::StyleText(wxStyledTextCtrl* ctrl, wxStyledTextEvent& e
...
             } else {
                 // File name
-                filenameStyleLen = 1;
+                filenameStyleLen = chWidth;
                 m_curstate = kFile;
             }
             break;
         case kLineNumber:
-            ++lineNumberStyleLen;
+            lineNumberStyleLen += chWidth;
```
eranif, line 166: why do you increase the line number by the char width here?

SFTP/SFTPGrepStyler.cpp
```diff
@@ -39,36 +43,36 @@ void SFTPGrepStyler::StyleText(wxStyledTextCtrl* ctrl, wxStyledTextEvent& e, boo
...
+            filenameStyleLen += chWidth;
             if(ch == ':') {
                 m_curstate = kLineNumber;
                 ctrl->SetStyling(filenameStyleLen, LEX_FIF_FILE);
                 filenameStyleLen = 0;
             }
             break;
         case kLineNumber:
-            ++lineNumberStyleLen;
+            lineNumberStyleLen += chWidth;
```
eranif, line 67: same comment as before.

Plugin/clFindResultsStyler.cpp
rlbxku1r: That's just to make it consistent look with other changes. The outcome should be same though (chWidth == 1), but I'll revert it back if it doesn't make sense.

eranif: Since we are advancing line numbers it makes no sense to make it depends on the char width (even though it will always be 1). For clarity, it should be `++lineNumberStyleLen`

rlbxku1r: Okay, reverted two of them.

SFTP/SFTPGrepStyler.cpp
rlbxku1r: Reverted (see another comment).

PR #2711, opened 2021-04-18
Fix FreeBSD13 build

rozhuk-im:
In file included from ../CodeLite/procutils.cpp:51:
In file included from /usr/include/sys/user.h:51:
In file included from /usr/include/sys/proc.h:48:
/usr/include/sys/condvar.h:40:22: error: reference to 'thread' is ambiguous
TAILQ_HEAD(cv_waitq, thread);


rozhuk-im:
@eranif ping

PR #2744, opened 2021-05-31
Patches that add or change to using "wx/msw/wrapwin.h"

stahta01:
Per an old wxWidgets doc file the use of "wx/msw/wrapwin.h" is preferred instead of including windows.h on the MS Windows OS. The use of Precompiled  headers is likely allowing the GUI to build without issues.
Without these changes I got these issues.
"#warning Please include winsock2.h before windows.h",
"error: 'SOCKET' does not name a type", or
"error: 'DWORD' does not name a type"
I replaced more uses of "windows.h" when I got an error in these three sections of code.

Tim S.


stahta01:
Found the directions [wxmsw-contributor-guide.md](https://github.com/wxWidgets/wxWidgets/blob/master/docs/contributing/wxmsw-contributor-guide.md) that says to use "wx/msw/wrapwin.h".

Tim S.

eranif:
FYI:
In order for this to work, on Windows, we should manually changing `wxUSE_SOCKET2` to `1` in `setup.h`, otherwise, the relevant block in `wrapwin.h` is ignored

PR #2753, opened 2021-06-09
Plugin: wxWidgets changes for MS Windows

stahta01:
Helps to prevent "error: 'ShowWindowAsync' was not declared" when building using cmake/make under MSys2

Added 2nd change to replace the included of "Winuser.h" because it caused build errors that suggested this header should not be included directly.


stahta01:
Add Changes to "codelite_terminal/wxTerminalColourHandler.cpp" to fix new build error.

PR #2768, opened 2021-06-24
Warning fix

stahta01:
Instead of "windows.h" or "winsock2.h" use "wx/msw/wrapwin.h".
Reason is to prevent warning message of
"Please include winsock2.h before windows.h"

And, I had to use wxS() to prevent a build error; not sure this patch is worth applying; but, I was always afraid that this warning might cause runtime issues. I have never read that it did; but, I assumed they added the warning for some good reason.

Tim S.


Jarod42:
Note: It seems you change all end of line. (didn't check the correct one).

stahta01:
Yes, I reformat the file as the owner of this project requested/requires!

Tim S.

PR #2782, opened 2021-07-21
Fix MSys2 Package build

stahta01:
I am not sure what your goal of adding the code I patched; so, it may or may not be correct.

Tim S.


eranif:
I would also like to be able to build CodeLite from command line (on native CMD terminal), without the need to download CodeLite first

PR #2785, opened 2021-07-26
cmake: Rewrite FindSqlite3.cmake and logic using it

stahta01:
I took your FindSqlite3.cmake and a tiny bit from the MSys2 and decided the logic and using the FindSqlite3 needed changed. I moved the part that prevents Apple from using the system Sqlite3 to the CMakeLists.txt file instead of in FindSqlite3.cmake.

Tim S.


stahta01:
I just added back the fatal error message for Unix builds.

Tim S.

PR #2827, opened 2021-09-09
Compiler: Tweak some default compiler options

rlbxku1r:
1. Make `-O*` descriptions consistent.
2. Update `-ansi` description according to the [current GCC manual](https://gcc.gnu.org/onlinedocs/gcc/C-Dialect-Options.html#index-ANSI-support).
3. Add `Enable C11 features` option which can be used to disable GNU C extensions.

This PR also contains a minor code refactoring for my earlier PRs.

### DIFF
--- a/Plugin/compiler.cpp
+++ b/Plugin/compiler.cpp
@@ -613,19 +613,18 @@ bool Compiler::IsGnuCompatibleCompiler() const
 void Compiler::AddDefaultGnuComplierOptions()
 {
     // Add GCC / CLANG default compiler options
-    AddCompilerOption("-O", "Optimize generated code. (for speed)");
-    AddCompilerOption("-O1", "Optimize more (for speed)");
-    AddCompilerOption("-O2", "Optimize even more (for speed)");
-    AddCompilerOption("-O3", "Optimize fully (for speed)");
-    AddCompilerOption("-Os", "Optimize generated code (for size)");
+    AddCompilerOption("-O", "Optimize generated code for speed");
+    AddCompilerOption("-O1", "Optimize more for speed");
+    AddCompilerOption("-O2", "Optimize even more for speed");
+    AddCompilerOption("-O3", "Optimize fully for speed");
+    AddCompilerOption("-Os", "Optimize generated code for size");
     AddCompilerOption("-O0", "Optimize for debugging");
     AddCompilerOption("-W", "Enable standard compiler warnings");
     AddCompilerOption("-Wall", "Enable all compiler warnings");
     AddCompilerOption("-Wfatal-errors", "Stop compiling after first error");
     AddCompilerOption("-Wmain", "Warn if main() is not conformant");
-    AddCompilerOption(
-        "-ansi",
-        "In C mode, support all ISO C90 programs. In C++ mode, remove GNU extensions that conflict with ISO C++");
+    AddCompilerOption("-ansi",
+                      "In C mode, this is equivalent to -std=c90. In C++ mode, it is equivalent to -std=c++98");
     AddCompilerOption("-fexpensive-optimizations", "Expensive optimizations");
     AddCompilerOption("-fopenmp", "Enable OpenMP (compilation)");
     AddCompilerOption("-g", "Produce debugging information");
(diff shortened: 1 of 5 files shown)


Plugin/compiler.cpp
```diff
@@ -613,7 +613,7 @@ bool Compiler::IsGnuCompatibleCompiler() const
 void Compiler::AddDefaultGnuComplierOptions()
 {
     // Add GCC / CLANG default compiler options
-    AddCompilerOption("-O", "Optimize generated code. (for speed)");
+    AddCompilerOption("-O", "Optimize generated code (for speed)");
```
Eric-01, line 616: Since you are touching it, what do you think about something like this?:
```cpp
    AddCompilerOption("-O", "Optimize generated code for speed");
    AddCompilerOption("-O1", "Optimize more for speed");
    AddCompilerOption("-O2", "Optimize even more for speed");
    AddCompilerOption("-O3", "Optimize fully for speed");
    AddCompilerOption("-Os", "Optimize generated code for size");
```

Usually text is put in parentheses when it is not important, and therefore can be safely omitted by the reader. However, in this case it is very important what the optimization target is.

rlbxku1r:
Pull Request updated:
1. Apply the suggestion mentioned by @Eric-01.
2. Update `-ansi` description according to the [current GCC manual](https://gcc.gnu.org/onlinedocs/gcc/C-Dialect-Options.html#index-ANSI-support).
3. Add `Enable ANSI C11 features` option which is primarily used to disable GCC extension (since it's already defaults to `-std=gnu17` in modern GCC).

Plugin/compiler.cpp
```diff
@@ -634,6 +633,7 @@ void Compiler::AddDefaultGnuComplierOptions()
     AddCompilerOption("-pg", "Profile code when executed");
     AddCompilerOption("-w", "Inhibit all warning messages");
     AddCompilerOption("-std=c99", "Enable ANSI C99 features");
+    AddCompilerOption("-std=c11", "Enable ANSI C11 features");
```
Eric-01, line 636: I think that only C89/C90 should be called "ANSI C", all later revisions just C.

Also, since you added C11 as a predefined compiler option, could you please add C17 as well?

rlbxku1r: > I think that only C89/C90 should be called "ANSI C", all later revisions just C.

Done, they're now `Enable C99/C11 features`.

> Also, since you added C11 as a predefined compiler option, could you please add C17 as well?

C17 is a minor revision containing bugfixes only. The [GCC manual](https://gcc.gnu.org/onlinedocs/gcc/C-Dialect-Options.html) states that those fixes are applied to C11 as well. So I'll leave it unchanged for now.

Eric-01: I have read the GCC manual as well, and you are right that since the C17 revision is so small, the GCC devs backported the changes to C11. However, GCC is not the only popular compiler oficially supported by CodeLite, and e.g. in case of Clang, I haven't found out whether the Clang devs also backported the changes from C17 to C11 or not. That's why I asked you the question in the first place.

Eric-01: I have read the GCC manual as well, and you are right that since the C17 revision is so small, the GCC devs backported the changes to C11. However, GCC is not the only popular compiler oficially supported by CodeLite, and e.g. in case of Clang, I haven't found out whether the Clang devs also backported the changes from C17 to C11 or not. That's why I asked you the question in the first place.

Oh, and thank you for your work, I really appreciate it :)

rlbxku1r: > I haven't found out whether the Clang devs also backported the changes from C17 to C11 or not.

I guess Clang does the same as GCC. I've looked at the Clang repository a bit and there were no differences except for `__STDC_VERSION__`.
Also, CodeLite still allows you to set custom compliler options, so the user can easily switch to C17 anytime if really needed.

Plugin/compiler.cpp
```diff
@@ -613,27 +613,27 @@ bool Compiler::IsGnuCompatibleCompiler() const
...
     AddCompilerOption("-pedantic", "Enable warnings demanded by strict ISO C and ISO C++");
     AddCompilerOption("-pedantic-errors", "Treat as errors the warnings demanded by strict ISO C and ISO C++");
     AddCompilerOption("-pg", "Profile code when executed");
     AddCompilerOption("-w", "Inhibit all warning messages");
-    AddCompilerOption("-std=c99", "Enable ANSI C99 features");
+    AddCompilerOption("-std=c99", "Enable C99 features");
+    AddCompilerOption("-std=c11", "Enable C11 features");
     AddCompilerOption("-std=c++11", "Enable C++11 features");
     AddCompilerOption("-std=c++14", "Enable C++14 features");
     AddCompilerOption("-std=c++17", "Enable C++17 features");
```
eranif, line 639: while you are at it, can you add `c++20`? thanks!

rlbxku1r: > while you are at it, can you add `c++20`? thanks!

It's already done in commit 6685f0d (was part of #2823).

PR #2872, opened 2021-11-20
LSP: Add support for `textDocument/hover`

rlbxku1r:
This PR makes `Display type info tooltips` to take the advantage of Language Server:
lsp-hover-fixed


eranif:
Nice one!
FYI: I am currently working on separating the built-in code completion (`ctags` based) into its own language server (`ctagsd`)
so this PR adds an additional work for me :D

PR #2889, opened 2021-12-14
Find & Replace: Regular expression fixes

rlbxku1r:
1. `Quick Find Bar`: Use POSIX-style syntax in regular expression mode.
Without [this flag](https://www.scintilla.org/SciTERegEx.html), `\(` `\)` are meant for capturing, whereas `(` `)` are for plain parentheses, which is exact opposite of more widely used syntax.
This change also make the behavior consistent with `Find in Files` search (which uses `wxRegEx` class).

2. `Find in Files`: Allow backref replace in regular expression mode.
`\number` (up to 9 captures) can be used as placeholder for captured strings.

Here's a typical example:
```c++
if(a == false) { /* ... */ }
if(bb == true) { /* ... */ }
if(ccc == false) { /* ... */ }
if(dddd == true) { /* ... */ }
if(eeeee == false) { /* ... */ }
if(ffffff == true) { /* ... */ }
```
the following replacement:
quickfindbar-regex
produces:
```c++
if(!a) { /* ... */ }
if(bb == true) { /* ... */ }
if(!ccc) { /* ... */ }
if(dddd == true) { /* ... */ }
if(!eeeee) { /* ... */ }
if(ffffff == true) { /* ... */ }
```

### DIFF
--- a/CodeLite/search_thread.cpp
+++ b/CodeLite/search_thread.cpp
@@ -382,6 +382,11 @@ void SearchThread::DoSearchLineRE(const wxString& line, const int lineNum, const
             result.SetLen(iCorrectedLen);
             result.SetFlags(data->m_flags);
             result.SetFindWhat(data->GetFindString());
+            wxArrayString regexCaptures;
+            for(size_t i = 0; i < re.GetMatchCount(); ++i) {
+                regexCaptures.Add(re.GetMatch(modLine, i));
+            }
+            result.SetRegexCaptures(regexCaptures);
 
             // Make sure our match is not on a comment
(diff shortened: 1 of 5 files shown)


eranif:
FYI:
Using the `Replace All` button corrupts the text (when using the placeholder)

rlbxku1r:
> FYI: Using the `Replace All` button corrupts the text (when using the placeholder)

Can you provide an example to reproduce the issue?

eranif:
I am unable to reproduce this now, but I will report back when I do manage to reproduce what I had

PR #2901, opened 2021-12-29
CYGWIN: avoid use of options for ELF targets.

carlo-bramini:
CYGWIN is recognized by CMake as UNIX, but it does not use ELF executables.
It uses the usual COFF-PE format.
The option "--disable-new-dtags" is unsupported because it can be appied only for ELF targets.
So, GCC hangs with an error when you try to link something on CYGWIN with that.
This PR suggests a tiny correction for fixing this error.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -82,7 +82,7 @@ if( UNIX AND NOT APPLE)
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -isystem /usr/include/harfbuzz") # Needed for fedora 31 so far, but will spread. See https://gitlab.kitware.com/cmake/cmake/issues/19531
 endif()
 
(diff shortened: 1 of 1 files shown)


eranif:
IIRC, I added this flag (i.e. `--disable-new-dtags`) couple years ago to re-enable `rpath` so we can install CodeLite libraries into `/usr/lib/x86_64-linux-gnu/codelite/` (for example) this way we can provide our own `wxWidgets` libraries (which at the time contained patches over the standard `wxWidgets`). On macOS we fix this by using `install_name_tool` and using `@executable_path`, how will this on CYGWIN? will `rpath` be working on CYGWIN without `--disable-new-dtags`?

eranif:
I am going to merge this, since it obvisouly does not work on CYGWIN as it is, so it can't do harm

PR #2910, opened 2021-12-31
codelite_indexer:  wxWidgets files must be included before others

carlo-bramini:
Similarly to PR #2906, wxWidgets files must be included first because they need to declare `_GNU_SOURCE` before using system headers.

### DIFF
--- a/sdk/codelite_indexer/utils.cpp
+++ b/sdk/codelite_indexer/utils.cpp
@@ -1,3 +1,4 @@
+// clang-format off
 #include <wx/arrstr.h>
 #include <wx/regex.h>
 #include <wx/string.h>
@@ -35,6 +36,7 @@
 #define strdup _strdup
 #endif
 #endif
+// clang-format on
 
 /**
  * helper string methods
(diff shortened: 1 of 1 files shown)


eranif:
Since I am using `clang-format` with this option: `IncludeBlocks: Regroup` i.e. it breaks the includes into 3 groups:
- If its a `cpp` then the header for that `cpp` will be placed first
- Second group contains "local" includes (includes using this format `#include ".."`)
- Global includes (includes using this format `#include <..>`

So I suggest that you wrap this change with:
```
// clang-format off
...
// clang-format on
```
this instructs `clang-foramt` to skip the code enclosed between these sections

carlo-bramini:
Excuse me, if I understood correctly, did you mean something like this:

```
// clang-format off
#include <wx/arrstr.h>
#include <wx/regex.h>
#include <wx/string.h>
#include <wx/tokenzr.h>
#include "utils.h"
#include "pptable.h"
#include <map>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
// clang-format on
```

?

eranif:
Yes. if you run `clang-format` on that file, it will skip this part.
otherwise it will sort the includes and will break your change

PR #2926, opened 2022-01-14
Revert "wxCrafter: Generate all getters as `noexcept` method"

### LINKED ISSUE #2925: MSVC 2013 error C3646: 'noexcept' : unknown override specifier
The Microsoft Compiler 2013 unfortunately does not like the noexcept specifiers in wxCrafter.h

Martin

rlbxku1r:
1. The introduction of `noexcept` keyword was a bit early, so revert 1966fbe.
fixes #2925.

2. Check `wxcWidget::KeepAsClassMember()` when generating getters.
This fix does nothing at the moment, but is also harmless and prevents an error from future wxCrafter changes.

### DIFF
--- a/wxcrafter/wxc_widget.cpp
+++ b/wxcrafter/wxc_widget.cpp
@@ -2735,8 +2735,7 @@ void wxcWidget::DoGenerateGetters(wxString& decl) const
                 firstChar.MakeUpper();
                 memberName.replace(0, 1, firstChar);
 
-                code << "    " << GetRealClassName() << "* Get" << memberName << "() noexcept { return " << GetName()
-                     << "; }\n";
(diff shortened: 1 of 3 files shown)


wxcrafter/wxc_widget.cpp
```diff
@@ -2696,7 +2696,7 @@ bool wxcWidget::IsParentAuiToolbar() const
 
 void wxcWidget::DoGenerateGetters(wxString& decl) const
 {
-    if(wxcSettings::Get().IsLicensed2()) {
+    if(wxcSettings::Get().IsLicensed2() && KeepAsClassMember()) {
```
eranif, line 2699: we can remove the `wxcSettings::Get().IsLicensed2()` its no longer use

rlbxku1r: Fixed.

PR #2981, opened 2022-04-09
Fixed wrong version for SEMIBOLD

pmconrad:
`SetFractionalPointSize` and `wxFONTWEIGHT_SEMIBOLD` were introduced in 3.1.2 not 3.1.0

### DIFF
--- a/Plugin/clButtonBase.cpp
+++ b/Plugin/clButtonBase.cpp
@@ -15,6 +15,8 @@
 #define TEXT_SPACER FromDIP(5)
 #else
 #define TEXT_SPACER 5
+#endif
+#if !wxCHECK_VERSION(3, 1, 2)
 #define SetFractionalPointSize SetPointSize
 #define wxFONTWEIGHT_SEMIBOLD wxFONTWEIGHT_BOLD
 #endif


Plugin/clButtonBase.cpp
```diff
@@ -15,6 +15,9 @@
 #define TEXT_SPACER FromDIP(5)
 #else
 #define TEXT_SPACER 5
+#endif
+#if wxCHECK_VERSION(3, 1, 2)
```
eranif, line 19: if you want to write: `if version >= 3.1.2` then you can simply use
```
#if wxVERSION_NUMBER >= 3120
..
#endif
```

pmconrad: Changed.

rlbxku1r: How about this:
```c
#if !wxCHECK_VERSION(3, 1, 2)
// ...
#endif
```
rather than:
```c
#if wxCHECK_VERSION(3, 1, 2)
#else
// ...
#endif
```
?

pmconrad: Wasn't sure about the ! operator in cpp expressions, but it seems to be working.
Dunno why my change didn't make it here, sorry.

PR #3005, opened 2022-05-27
Revert all changes to unicode string literals

rlbxku1r:
This is another fix for the commit 5a99fa1 (which was introduced in my earlier commit 26312f0).


eranif:
I already reverted some of those ... I wondered why **I** removed the `wxT(..)` (these symbols are not meant to be translated or we will lose the drawing) ... now I know how this happened :)

PR #3020, opened 2022-06-07
Allow external RPATH additions

uffejakobsen:
With the recent requirement for unstable wx-3.1.x - distribution packagers needs to overload the hardcoded RPATH and (along with the plugin path) append a path for wx-unstable to RPATH


uffejakobsen:
Any problems with this PR ?

On archlinux the unstable/unofficial wx-3.1.x resides in /opt/wxgtk-dev/ in order to co-exist with the official wx-3.0.x in /usr

The extra library path for the unstable wx-3.1.x needs to be injected into the codelite binaries through RPATH - along with the existing paths for other codelite libraries (/usr/lib/codelite/ etc)

If omitted starting codelite from an desktop-icon will not work - due to codelite missing an additional library path for wx-3.1.x

uffejakobsen:
thanks :+1:

PR #3042, opened 2022-07-23
Fix build/compile error on linux

uffejakobsen:
My guess is that your primary development platform is Windows ?
What became of the github actions workflow builds - it would have found errors like this one.


eranif:
actually, I always compile on Linux... for some reason, it never failed for me

PR #3122, opened 2023-01-03
libdatabaselayer.so installs in wrong location

uffejakobsen:


eranif:
this will break Windows (we are now using cmake on Windows as well)
I will push a commit to fix this soon

PR #3124, opened 2023-01-03
Fix wrong install location for templete files

uffejakobsen:


eranif:
thanks for reporting this. the correct fix should be using the `CL_RESOURCES_DIR` variable (your fix would have break MinGW)
I committed a fix for this

PR #3125, opened 2023-01-03
libdapcxx.so installs in wrong location

uffejakobsen:


eranif:
I have added new macro `codelite_install_library_target` and committed a fix for this

PR #3288, opened 2023-12-02
Update download.md to remove apt-key usage

jgmoss:
Remove apt-key due to it being deprecated. Replaced it with recommended method.


Tiacameron:
Okay

Tiacameron:
jgmoss:master

PR #3300, opened 2024-01-27
Get back toolbar buttons and add few new buttons

rozhuk-im:
Return toolbar buttons deleted by https://github.com/eranif/codelite/commit/301011dbd196fc73cfda469424269646924fed0b

Please merge this, I do not want remember multiple hotkeys, and CTRL+SHIFT+* will not work for me, because I use CTRL+SHIFT to switch languages.


rozhuk-im:
Thanks!

Issue #3298, opened 2024-01-27
[Bug]: Codelite hand during rebuild project


rozhuk-im:
### What happened?

Codelite infinite call ```BuildTab::OnBuildAddLine``` have no idea why: I try to debug this but not found who post events, probably it try to process same event infinite times.
If it rely on ```ZombieReaperPOSIX::Entry()``` - this can be reason, because this code may lost events IMHO.

OS: FreeBSD 14/stable

### Version

Self compiled

### Operating system

Other

### Steps to reproduce

```text
git clone https://github.com/rozhuk-im/gtk-mixer.git
Open workspace in Codelite
Try to REBUILD debug configuration
```

rozhuk-im:
Looks like I found why this happen.

Project generate 30+mb log and 500k+ lines of output.

```
void clRowEntry::InsertChild(clRowEntry* child, clRowEntry* prev)
....
    // Connect the linked list for sequential iteration
    clRowEntry::Vec_t::iterator iterCur =
        std::find_if(m_children.begin(), m_children.end(), [&](clRowEntry* c) { return c == child; });
....
```
This become very slow after 100k items.

Also ````lines.erase(lines.begin());``` looks slow and can be replaced with:
```
    auto lines = ::wxStringTokenize(m_buffer, "\n", wxTOKEN_RET_DELIMS);
    wxString remainder;
    //while(!lines.empty()) {
    const size_t cnt = lines.Count();
    for (size_t i = 0; i < cnt; i ++) {
        auto& line = lines[i];
```

eranif:
Thanks, looking forward to apply your PR. I posted 2 comments on it (the first comment is crucial, the second is more of a "nit" comment)

rozhuk-im:
https://github.com/eranif/codelite/pull/3304 and https://github.com/eranif/codelite/pull/3305 also here.
Current limit with 8192 lines mostly unreachable, now limiting  call```m_view->Commit();``` and ```m_view->ScrollToBottom();``` save more CPU time than limiting for lines parse.

eranif:
> ....
>    // Connect the linked list for sequential iteration
>   clRowEntry::Vec_t::iterator iterCur =
>        std::find_if(m_children.begin(), m_children.end(), [&](clRowEntry* c) { return c == child; });
>....

This looks redundant. We could use the previous insert and capture the iterator

So the code above this snippet looks like:

```c++
    // We need the last item of this subtree (prev 'this' is the root)
[...]

PR #3303, opened 2024-01-28
Fix GUI hang on build in case too big output

rozhuk-im:
Fix for https://github.com/eranif/codelite/issues/3298

Added process rate limit because ```m_view->Commit();``` and ```m_view->ScrollToBottom();``` heavy. (100 ms)
Added output lines count process limit to reduce ```m_activeCompiler->Matches()``` calls (8192 lines)

Improve ```clRowEntry::DeleteAllChildren()``` and ```clRowEntry::InsertChild()``` speed.

### DIFF
--- a/LiteEditor/BuildTab.cpp
+++ b/LiteEditor/BuildTab.cpp
@@ -57,6 +57,8 @@ BuildTab::BuildTab(wxWindow* parent)
 
     wxTheApp->Bind(wxEVT_MENU, &BuildTab::OnNextBuildError, this, XRCID("next_build_error"));
     wxTheApp->Bind(wxEVT_UPDATE_UI, &BuildTab::OnNextBuildErrorUI, this, XRCID("next_build_error"));
+
+    m_buffer_sw.Start();
 }
 
 BuildTab::~BuildTab()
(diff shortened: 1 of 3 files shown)


LiteEditor/BuildTab.cpp
```diff
@@ -144,12 +146,27 @@ void BuildTab::OnBuildEnded(clBuildEvent& e)
 
 void BuildTab::ProcessBuffer(bool last_line)
 {
+    // Limit data process frequency.
+    if(!last_line) {
+        uint64_t time_diff_ms;
+        struct timespec cur_time;
+
+        clock_gettime(CLOCK_MONOTONIC, &cur_time);
```
eranif, line 154: this won't compile on Windows. If you are looking into applying some throttling and looking for high resolution clock, consider using `wxStopWatch` or use `std::chrono::*`

LiteEditor/BuildTab.cpp
```diff
@@ -174,6 +191,10 @@ void BuildTab::ProcessBuffer(bool last_line)
             // the internal project
             m_view->AppendItem(line);
 
+         } else if(cnt > 8192) {
```
eranif, line 194: can we move this to be constant?

PR #3305, opened 2024-01-30
Switch to more faster regexps for gcc/clang

rozhuk-im:
This dramatically improve compiler output lines parsing.
May be some thing missed, but most useful messages handled.


rozhuk-im:
regexp from: https://egeek.me/2015/03/20/jenkins-custom-clang-warnings-parser/

eranif:
My main problem with this PR is that these regex expressions were built over the years and they cover, well, I think - everything.
Your changes might be faster (actually, I am sure they are..) but there were not tested. I putting this PR on hold until I release CodeLite 18.0.0 - hopefully, this weekend

PR #3337, opened 2024-03-19
[CI] Add github action for macos

Jarod42:
52m job https://github.com/Jarod42/codelite/actions/runs/8343196567
spot the issue about SideBarToolBar :) https://github.com/Jarod42/codelite/actions/runs/8339588463/job/22821858878

[edited]: only 25min here :-)

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -1,0 +1,50 @@
+name: macos
+
+on:
+  workflow_dispatch:
+  push:
+  pull_request:
+
+jobs:
+  macos:
+    runs-on: macos-latest
+
+    steps:
+
+    - name: Install Codelite's dependencies
+      run: |
+        # brew update && brew upgrade # fails somehow
+        brew install git cmake libssh hunspell
+
+    # WxWidgets
+    - name: Checkout
+      uses: actions/checkout@v4
+      with:
+          repository: wxWidgets/wxWidgets
+          path: wxWidgets
+          submodules: recursive
+    
+    - name: Build and install wxWidgets
+      run: |
+        mkdir wxWidgets/build-release
+        cd wxWidgets/build-release
+        ../configure --enable-shared --enable-monolithic --with-osx_cocoa CXX='clang++ -std=c++11 -stdlib=libc++ -I../src/tiff/libtiff' CC=clang --disable-debug --disable-mediactrl --enable-stl
+        make -j10
+        sudo make install
+
+    # Codelite
+    - name: Checkout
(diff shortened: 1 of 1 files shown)


.github/workflows/macos.yml
```diff
@@ -0,0 +1,50 @@
...
+    - name: Build and install Codelite
+      run: |
+        mkdir build-release
+        cd build-release
+        cmake .. -DCMAKE_BUILD_TYPE=Release
+        cmake --build . -j $(sysctl -n hw.physicalcpu)
+        cmake --build . --target install
+
+    - name: codelite --version
+      run: codelite --version || exit 0 # codelite --version returns -1
```
eranif, line 50: Does this work? it should be:

```bash
codelite.app/Contents/MacOS/codelite --version
```

Jarod42: > /Users/runner/work/_temp/02ffa54b-035d-4b75-bef9-f8728bd7b170.sh: line 1: codelite: command not found

Indeed, not.
Changed with your version, waiting for CI.

Jarod42:
> codelite.app/Contents/MacOS/codelite: No such file or directory

That version doesn't work neither

Whole block can be removed (As important think, the build, passes).
Or macos users might check correct command (I don't know if it is path issue, or something similar).

.github/workflows/macos.yml
```diff
@@ -0,0 +1,50 @@
...
+          repository: wxWidgets/wxWidgets
+          path: wxWidgets
+          submodules: recursive
+    
+    - name: Build and install wxWidgets
+      run: |
+        mkdir wxWidgets/build-release
+        cd wxWidgets/build-release
+        ../configure --enable-shared --enable-monolithic --with-osx_cocoa CXX='clang++ -std=c++11 -stdlib=libc++ -I../src/tiff/libtiff' CC=clang --disable-debug --disable-mediactrl --enable-stl
+        make -j10
```
Eric-01, line 32: Why hardcode it to 10 jobs?

Also, CodeLite is now built using C++17, no? Is there any reason why wxWidgets is built using an older standard on macOS (I've never owned a Mac)?

eranif: @Jarod42 we could probably use

```bash
 make -j$(sysctl -n hw.ncpu) install
``

Jarod42: I mostly copy-paste code from https://docs.codelite.org/build/build_wx_widgets/#macos

I don't have MacOS neither.

Created https://github.com/eranif/codelite/pull/3339.

PR #3344, opened 2024-04-09
Fix SIOF (Static initializer order Fiasco) 

Jarod42:
with lazy initialization of loggers:

`clStandardPaths::Get().GetUserDataDir()` (called from Logger constructor) requires that Application is set.


Jarod42:
ctags' build seems unstable (don't pass here on msys2, CI passes on my repo https://github.com/Jarod42/codelite/actions/runs/8615923100/job/23612457688 ...

PR #3350, opened 2024-04-15
Remove old remaining sqlite3 references.

### LINKED ISSUE #2952: Multiple local copies of SQLite3
I have seen that there are several copies of amalgamed SQLite3 into the code tree:

```
codelite\codelite-cli
version 3.24.0
size ~7.4MB
```

```
sdk\databaselayer\src\sqlite3
version 3.7.12.1
size ~4.8MB
```

```
sqlite3
version 3.7.12.1
size ~4.8MB
```

Since I'm using the system library, these local copies do not affect my build (I hope), but perhaps it would be useful to use just one of them as static library and delete the others, for reducing a bit the size of the source archive.

Jarod42:
closes #2952


eranif:
Merging this, but FYI: I have stopped using these .project files 2 or more years ago

Jarod42:
I was unsure if those projects can be removed.
only root ones are used?
Parser ones too? (the ones which copy generated file from lex|flex, yacc|bison)?

PR #3352, opened 2024-04-18
Flex cmake

Jarod42:
Build with flex, and remove generated file from repo.
Restore CodeLite/include_finder.l
Report build fixes from generated file CodeLite/PhpLexer.cpp to source file CodeLite/PhpLexer.l


UffeJakobsen:
Question - I may have missed some detail here - but wont this open up for the risk of different versions of flex producing different output ?
Ultimately resulting in weird problems on different distributions that is difficult to debug

Jarod42:
@UffeJakobsen: 
The different output should be equivalent, as for gcc/clang/cmake version used...
Chances are than newer versions are even better.
That avoids "desynchronization" between sources and generated files (which might be source of bug too). (Yes, some of the generated files has been changed manually :-/)

If needed, I might changed to still put generated files in source tree, but I prefer to not have generated files in source tree.
(I plan to also remove generated files from wxcrafter in the future, but it requires some works in the tool first).

UffeJakobsen:
Well at my $WORK-place - some 6-7 years ago - we had an obscure problem - some of our binary releases worked different than others...
The problem was traced down to differences in the flex version that was used to build the release.
That was before we got real CI/CD systems - the sw-engineers that did the release builds each had their own favorite system on which they did the release builds... and at least one of these systems was running a distro that had a flaky flex....

Jarod42:
What is really different between Flex than other dependencies (libssh, wxwidgets, hungspell, ...) which might have different version?
We might still require a specific version from CMake IIRC.

AJenbo:
@Jarod42 Would have been hard for me to make the recent fixes to the PHP parser without this clean up, so thank you.

PR #3370, opened 2024-05-24
CxxParser compile fixes for linux etc

UffeJakobsen:


Jarod42:
What is missing from CI to spot the issue?
(I don't have nix/apple to test locally :-/ so sorry for the break).

Issue #3396, opened 2024-06-03
[Bug]: Infinit indexing on looping symlinks


AJenbo:
### What happened?

Some packages (like `orchestra/testbench`) may symlink the vendor folder, causing an infinitly nested folder structure, leaving CodeLite stuck trying to analyze the files and eventually consuming all memory.

I'm not 100% sure but I think this happens somewhere in the parser since it appears to finish the "scanning for workspace files" step before getting stuck, though the file list never appears in the workspace panel.

### Version

Other

### Operating system

Linux

### Relevant log output

```shell
17.0.0 & 17.10.0
```

PR #3401, opened 2024-06-07
Check for looping symlinks when building PHP projects

AJenbo:
Fixes https://github.com/eranif/codelite/issues/3396

Finally located the issue for the IDE getting stuck on some project, this has brought so much pain over the last few years :sob: 
(I had gotten used to removing the vendor folder before starting CodeLite, or loading a different project and adding search paths)

The solution here is to resolve each folder before scanning it and then keeping track of the resolved paths, if we encounter a real path we have already seen previously we do not visit the symlink.
The alternative could be to simply not follow symlinks, but this feels like a better solution as it supports cases where the symlinks are relevant, but avoids cases where they are self nesting.

### DIFF
--- a/codelitephp/PHPParser/FilesCollector.cpp
+++ b/codelitephp/PHPParser/FilesCollector.cpp
@@ -52,8 +52,7 @@ void FilesCollector::Collect(const wxString& rootFolder)
             bool isDirectory = wxFileName::DirExists(fullpath);
             if(isDirectory && (m_excludeFolders.count(filename) == 0)) {
                 wxString canonicalPath = wxFileName(fullpath).ResolveLink().GetFullPath();
-                if (visitedFolders.find(canonicalPath) == visitedFolders.end()) {
-                    visitedFolders.insert(canonicalPath);
+                if (visitedFolders.insert(canonicalPath).second) {
                     // A directory
                     Q.push(fullpath);
                     fullpath << wxFileName::GetPathSeparator() << FOLDER_MARKER;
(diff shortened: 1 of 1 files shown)


eranif:
Actually, a more correct fix would be to switch to the general purpose `clFilesScanner` (in file `clFilesCollector.cpp`) class which already keep tracks of visited folders and resolves symbolic links

codelitephp/PHPParser/FilesCollector.cpp
```diff
@@ -50,11 +51,14 @@ void FilesCollector::Collect(const wxString& rootFolder)
...
             bool isDirectory = wxFileName::DirExists(fullpath);
             if(isDirectory && (m_excludeFolders.count(filename) == 0)) {
-                // A directory
-                Q.push(fullpath);
-                fullpath << wxFileName::GetPathSeparator() << FOLDER_MARKER;
-                V.push_back(fullpath);
-
+                wxString canonicalPath = wxFileName(fullpath).ResolveLink().GetFullPath();
+                if (visitedFolders.find(canonicalPath) == visitedFolders.end()) {
+                    visitedFolders.insert(canonicalPath);
```
Jarod42, line 56:
```suggestion
                if (visitedFolders.insert(canonicalPath).second) {
```
or
```suggestion
                if (const auto [it, inserted] = visitedFolders.insert(canonicalPath); inserted) {
```

To do only one lookup.

AJenbo:
> Actually, a more correct fix would be to switch to the general purpose `clFilesScanner` (in file `clFilesCollector.cpp`) class which already keep tracks of visited folders and resolves symbolic links

I took a quick stab at this, one thing I noticed is that `clFilesScanner` does not have support for empty folders. Another thing that I noticed is that the progress bar stopped around 33% after the switch, but I'm not sure why that is, could just be my implementation.

I'm not sure what the best way to add folder support to the general scanner would be... and I still have a few modern PHP features that I would like to focus on adding support for first :)

PR #3419, opened 2024-07-05
Fix warning about `printf`:

Jarod42:
- warning: format ‘%d’ expects argument of type ‘int’, but argument 2 has type `std::size_t`.

### DIFF
--- a/CxxParser/main.cpp
+++ b/CxxParser/main.cpp
@@ -68,7 +68,7 @@ void testFuncParser(char *buf)
     }
 
 //	printf("total time: %d\n", end-start);
-    printf("matches found: %d\n", li.size());
+    printf("matches found: %d\n", (int)li.size());
 }
 
 void testExprParser(char *buf)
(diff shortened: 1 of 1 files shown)


Eric-01:
Why not `%zu` instead?

Eric-01:
Why not `%zu` instead? :)

Jarod42:
Not familiar enough with `printf` format. I know that `std::size_t` might be of different size, warning suggests `%ld` as fix, so wrong for other size. Not sure neither if `%zu` is supported by all compiler...

AJenbo:
Testing via [GodBolt](https://godbolt.org/z/W7fjr75c5) it seems it should be supported on:
Clang 5, GCC 5, djgpp 5.5, icc 16, icx 2021, MSVC 19, nvc++ 22, FRC, TI CL430, Zapcc

It appears that it's more common for compilers to not support C++17 then `%zu`.

PR #3423, opened 2024-07-12
Plugin sub directories

Jarod42:


Jarod42:
Rebased to also apply changes to new code

PR #3445, opened 2024-08-05
codelite-official cask does not exist on homebrew anymore

DDAN-17:


eranif:
Yes it does... `codelite-official` is our cask. You need to tap it before you can use it:

```
brew tap eranif/codelite
```

The repository is here:
https://github.com/eranif/homebrew-codelite

PR #3446, opened 2024-08-09
[CI] Fix msys CI and use both wxWidgets 'v3.2.5' and 'master'.

Jarod42:
wxWidgets sha1 3737245c5195d84f0c788e650c6ad7992eec03d7 no longer works with Codelite Fix


Eric-01:
Maybe you could do the same for Linux and macOS?

eranif:
Thanks for fixing this

Issue #3451, opened 2024-08-14
[Bug]: Building Codelite from Sources fails to build


noeldiviney:
### What happened?

Building Codelite From sources as described in "https://docs.codelite.org/build/build_fr
om_sources" has new issue 

[BuildCodeliteFromSourcesFailsLog.txt](https://github.com/user-attachments/files/16612598/BuildCodeliteFromSourcesFailsLog.txt)

### Version

Self compiled

### Operating system

Windows 10

### Steps to reproduce

Building Codelite From sources as described [here](https://docs.codelite.org/build/build_from_sources)

[BuildCodeliteFromSourcesFailsLog.txt](https://github.com/user-attachments/files/16612598/BuildCodeliteFromSourcesFailsLog.txt)

Noel Diviney

### Relevant log output

[BuildCodeliteFromSourcesFailsLog.txt](https://github.com/user-attachments/files/16612598/BuildCodeliteFromSourcesFailsLog.txt)

noeldiviney:
Successful build of Codelite Sources on both Windows 10 MSYS2 and Linux Mint 22-04.
Many thanks Eran.
Best Regards Noel.

PR #3452, opened 2024-08-16
[build] Hack to fix msys2 build

Jarod42:
I don't know why CI suddenly failed on that :-/

Proper fix would probably require to check version of MinGW (function proptotype seems to be present from at least 5 years...)
Dropping the check would be an alternative.


eranif:
Related to https://github.com/eranif/codelite/issues/3451

eranif:
With this commit a12c8b88895043aacde2940daad6a068a307fd61, this is no longer needed

PR #3454, opened 2024-08-20
Update v18_0_0.md

AJenbo:
Felt like it would be worth mentioning these as I could see them as being critical for some users


eranif:
Thanks merged. FYI: I usually update this file before releasing a new version

PR #3471, opened 2024-09-17
[cleanup]:

Jarod42:
- remove commented code and empty functions.
- use `explicit`
- add `override`
- replace output parameter by return type
- replace `NULL` by `nullptr`
- replace `std::for_each` by for-range
- fix typo (indenation -> indentation; overrite -> overwrite)
- clang-format


AJenbo:
Nice work

PR #3481, opened 2024-09-24
Thanks to MSVC warning, fix `DoReadProjectTemplatesFromFolder`

Jarod42:
MSVC warns about shadowing of `files`
```
    wxArrayString files;
    if (wxFileName::DirExists(folder)) {
       // ...
        const auto& files = dt.GetFiles();
```
so the outer is never set.

### DIFF
--- a/Plugin/globals.cpp
+++ b/Plugin/globals.cpp
@@ -730,7 +730,6 @@ static void DoReadProjectTemplatesFromFolder(const wxString& folder, std::list<P
                                              bool loadDefaults = true)
 {
     // read all files under this directory
-    wxArrayString files;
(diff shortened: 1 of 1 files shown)


Plugin/globals.cpp
```diff
@@ -761,7 +760,7 @@ static void DoReadProjectTemplatesFromFolder(const wxString& folder, std::list<P
         }
     }
 
-    if (loadDefaults && files.IsEmpty()) {
+    if (loadDefaults && list.empty()) {
```
Jarod42, line 763: Previously, files was always empty.
I suspect you wanted to check `list` from below comment.
For previous behavior, remove ` && files.IsEmpty()` instead.

PR #3510, opened 2024-10-23
[CMake][CI] Allow to build DatabaseExplorer plugin with mariaDb (mysql)

Jarod42:
Compile DatabaseExplorer with mysql (mariaDb) enabled by default (but MacOS which default to OFF).

Note: I cannot test behavior for Ubuntu/MacOS.
I did minimal checks on minGW (connect to empty DB).

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -44,7 +44,7 @@ jobs:
         mkdir build-release
         cd build-release
         brew install autoconf automake libtool gettext
-        cmake .. -DCMAKE_BUILD_TYPE=Release -Wno-dev -DBUILD_TESTING=1
+        cmake .. -DCMAKE_BUILD_TYPE=Release -Wno-dev -DWITH_MYSQL=1 -DBUILD_TESTING=1
         cmake --build . -j $(sysctl -n hw.physicalcpu)
         cmake --build . --target install
         ctest --output-on-failure
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -14,7 +14,7 @@ jobs:
     - name: Install Codelite's dependencies
       run: |
         # brew update && brew upgrade # fails somehow
-        brew install git cmake libssh hunspell bison flex
+        brew install git cmake libssh hunspell bison flex mariadb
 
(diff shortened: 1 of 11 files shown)


eifrah-aws:
The problem with macOS is not installing the libraries is actually the packaging process. Atm, when compiling and packaging for macOS, `/opt/homebrew` is not in the PATH (on purpose) - no all mac users are using `brew` (some are using `The MacPorts Project`). So we can't really install binaries from `/opt/homebrew`. My suggestion: make this optional on macOS until I will figure the packaging stuff on macOS

eranif:
The problem with macOS is not installing the libraries is actually the packaging process. Atm, when compiling and packaging for macOS, `/opt/homebrew` is not in the PATH (on purpose) - no all mac users are using `brew` (some are using `The MacPorts Project`). So we can't really install binaries from `/opt/homebrew`. My suggestion: make this optional on macOS until I will figure the packaging stuff on macOS

Jarod42:
Unsure to understand what you mean,
with `option(WITH_MYSQL "Enable support of MySQL for DatabaseExplorer plugin" ON)`
user might disable it...

eifrah-aws:
> Unsure to understand what you mean, with `option(WITH_MYSQL "Enable support of MySQL for DatabaseExplorer plugin" ON)` user might disable it...

The default should be `OFF`.
we could also change it to:


```cmake
if (APPLE)
    option(WITH_MYSQL "Enable support of MySQL for DatabaseExplorer plugin" OFF)
else ()
    option(WITH_MYSQL "Enable support of MySQL for DatabaseExplorer plugin" ON)
endif ()
```

eranif:
> Unsure to understand what you mean, with `option(WITH_MYSQL "Enable support of MySQL for DatabaseExplorer plugin" ON)` user might disable it...

The default should be `OFF`.
we could also change it to:


```cmake
if (APPLE)
    option(WITH_MYSQL "Enable support of MySQL for DatabaseExplorer plugin" OFF)
else ()
    option(WITH_MYSQL "Enable support of MySQL for DatabaseExplorer plugin" ON)
endif ()
```

PR #3525, opened 2024-11-05
[cleanup] Fix some English typos

Jarod42:


CodeLite/clINIParser.cpp
```diff
@@ -27,7 +27,7 @@ void clINISection::AddKeyValue(const wxString& key, const wxString& value)
     m_entries.push_back(kv);
 }
 
-void clINISection::Serialise(std::ostream& string) const
```
eranif, line 30: This one is actually the English way of writing it "Serialise", "Initialise" 😉 , some will argue that your fix is the wrong one (me included)

eranif (approved):

CodeLite/clINIParser.cpp
Jarod42: :)
All others places use `Serialize`.
Didn't check Initiali[sz]e.

PR #3531, opened 2024-11-13
Removed font size multiplier

dee:
On Windows, IDE fonts in a project tree, editor tabs etc are noticeably larger than a system font. That was because an IDE font was 20% bigger than a system one. Tested it on 4K and 1080p.

### DIFF
--- a/Plugin/clScrolledPanel.cpp
+++ b/Plugin/clScrolledPanel.cpp
@@ -384,7 +384,7 @@ wxFont clScrolledPanel::GetDefaultFont()
     f.SetFractionalPointSize(pointSize);
 #elif defined(__WXMSW__)
     // wxMSW
-    float pointSize = f.GetFractionalPointSize() * 1.2;
+    float pointSize = f.GetFractionalPointSize();
     f.SetFractionalPointSize(pointSize);
 #endif
     return f;
--- a/Plugin/clScrolledPanel.cpp
+++ b/Plugin/clScrolledPanel.cpp
@@ -379,11 +379,9 @@ void clScrolledPanel::DoCancelDrag()
 wxFont clScrolledPanel::GetDefaultFont()
 {
     wxFont f = wxSystemSettings::GetFont(wxSYS_DEFAULT_GUI_FONT);
-    float pointSize = f.GetFractionalPointSize();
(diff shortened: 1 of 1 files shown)


Plugin/clScrolledPanel.cpp
```diff
@@ -384,7 +384,7 @@ wxFont clScrolledPanel::GetDefaultFont()
     f.SetFractionalPointSize(pointSize);
 #elif defined(__WXMSW__)
     // wxMSW
-    float pointSize = f.GetFractionalPointSize() * 1.2;
+    float pointSize = f.GetFractionalPointSize();
     f.SetFractionalPointSize(pointSize);
```
Jarod42, line 388: So whole `__WXMSW__` block can be removed

dee: Something like this, i guess.

Jarod42: No need to set same value for other, simply
```
wxFont clScrolledPanel::GetDefaultFont()
{
    wxFont f = wxSystemSettings::GetFont(wxSYS_DEFAULT_GUI_FONT);
#if defined(__WXMAC__)
    float pointSize = f.GetFractionalPointSize() * 1.2;
    f.SetFractionalPointSize(pointSize);
#endif
    return f;
}
```
or
```
wxFont clScrolledPanel::GetDefaultFont()
{
    wxFont f = wxSystemSettings::GetFont(wxSYS_DEFAULT_GUI_FONT);
#if defined(__WXMAC__)
    f.SetFractionalPointSize(1.2 * f.GetFractionalPointSize());
#endif
    return f;
}
```

dee: Yes, really, I am setting the same value back.

Jarod42 (approved):

dee:
Thank you!

Jarod42:
Unsure why you close your PR.

Note that:
- approving != merging.
- I don't have write access, I'm just a contributor like you.

dee:
My bad, I thought it was finished.

eranif:
Even though some CI are failing - these were fixed in master branch. Once the Ubuntu build is OK, I will merge it

PR #3534, opened 2024-11-16
[CI] "Fix" CI for msys, wxWidget doesn't compile with clang-19 because …

Jarod42:
…of an very old bug in template.

It has been fixed by d275dd4ba8c07cf1b05077ebf2c1c2e67655d633 by lauri Nurmi the Oct 28th 2024 (for 3.2.7) See https://github.com/wxWidgets/wxWidgets/pull/24922 for more details


eranif:
FYI: this commit (which I pushed couple of hours ago): 4dc3bcddbec9d2cb6eb414a19dd880156f535ad4
also fixed it (I simply removed the `wxUSE_STL` from the cmake build)

eranif:
So if we are going to push your commit, we should probably revert my commit

Jarod42:
b9918c156b5cef50d0536670d522a41f9009b05f have to be reverted once https://github.com/wxWidgets/wxWidgets/issues/24958 is fixed

PR #3537, opened 2024-11-19
[cleanup] Code removal of unused code

Jarod42:


eranif:
Out of curiosity, are you using some kind of a tool to detect dead code?

Jarod42:
I now use my own tool [ccccc](https://github.com/Jarod42/ccccc) (based on llvm tooling) which collects generic metrics but also caller/call count.
I also have some "issues" with template/virtual methods (and also `main` :) ) to decide and have correct count.
but except that, I got accurate stats for regular functions never called.

I create a github workflow https://github.com/Jarod42/codelite/blob/ccccc/.github/workflows/ccccc.yml to collect the data (not entirely automated, I do manual changes afterward to fix internal path)

In my TODO list, I planned to
- split the dead-code detection in dedicated tool
- integrate my tools as codelite plugin :)

Jarod42:
For info,
I migrated workflow to ccccc directly https://github.com/Jarod42/ccccc/actions/workflows/demo-codelite.yml
Result is at https://jarod42.github.io/ccccc/codelite-index.html

PR #3540, opened 2024-11-23
[cleanup] Database explorer dead code removal

Jarod42:
Care to check icon changes.
Did I miss
- bitmaps/16-database.png
- bitmaps/16-database@2x.png
- bitmaps/24-database.png
- bitmaps/24-database@2x.png


eranif:
This is the old format I believe. Theses days (for couple of years now) CodeLite uses SVG images. We don't use bitmaps directly (unless I missed some). We use SVG which are adjustable in Runtime

PR #3544, opened 2024-12-01
Add new macro `WorkspaceRealPath`

UffeJakobsen:
Add new macro `WorkspaceRealPath`

The `WorkspaceRealPath` macro is identical to `WorkspacePath` - but just expanded/converted by calling FileUtils::RealPath (`realpath()`) - this means that any symlink present in the path is resolved.

Reasoning for introduction of `WorkspaceRealPath` macro:

I have a number of (FSW) projects that have to be placed on a  network drive with a special and extremely long/deep directory layout (not my choice)

In order to eliminate the long/deep path leading to the project dirs of these projects - I have created a local project root with symlinks to the real projects - mission accomplished...

This has worked well with CodeLite until the switch from ctags to clangd (LSP)...

Here is an example of one the problems (there are actually more):

Clangd (LSP) returns filepaths as expanded (real)paths.

With a FSW project opened with a shortend (symlink) path - the Workspace pane file-tree does not recognize the expanded (real)path as a part of the FSW project - and hence it will not sync the file returned by LSP with the workspace pane file-tree.

But clangd (LSP) has a fix for that - `path-mappings` will translate a remote path into a local path (just what we (I) need)

By adding the following into the clangd (LSP) command every thing works again:

`--path-mappings=$(WorkspacePath)=$(WorkspaceRealPath)`

Summary:

By introducing the `WorkspaceRealPath` macro the problems of mapping between "virtual" symlink paths and (real)paths are automatically handled

### DIFF
--- a/Plugin/macromanager.cpp
+++ b/Plugin/macromanager.cpp
@@ -35,6 +35,7 @@
 #include "project.h"
 #include "workspace.h"
 #include "wxStringHash.h"
+#include "fileutils.h"
 
 #include <wx/regex.h>
 
@@ -156,6 +157,7 @@ const std::unordered_set<wxString> CODELITE_MACROS = {
     "WorkspaceName",
     "WorkspaceConfiguration",
     "WorkspacePath",
+    "WorkspaceRealPath",
     "OutputDirectory",
     "ProjectOutputFile",
     "OutputFile",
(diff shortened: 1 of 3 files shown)


eranif:
Nice! I have been struggling with similar issue myself :) (I am using FSW with SSH)
At work, my home drive is `/local/home/...` so `clangd` resolves all the path to their real paths, so i might end up with 2 files opened at CodeLite the one with realpath and the symlinked, sounds like your fix will resolve both issues

UffeJakobsen:
Thanks for merging :+1: 

BTW: I have a few other patches that may be needed to cover all scenarios editor ends up opening 2 tabs for the same file... but they need a little clean up before they can become into PR's :-)

PR #3549, opened 2024-12-15
+ Remote debugging for Filesystem workspace

iHardRock:
Remote debugging now available in FS workspace as well.

изображение

Also bug with Program path on General  tab in FS workspace fixed. It was may be self-cleared after doing some changes on other tabs and then save not returning to General tab.

### DIFF
--- a/Plugin/FileSystemWorkspace/FSConfigPage.cpp
+++ b/Plugin/FileSystemWorkspace/FSConfigPage.cpp
@@ -32,6 +32,7 @@ FSConfigPage::FSConfigPage(wxWindow* parent, clFileSystemWorkspaceConfig::Ptr_t 
         lexer->Apply(m_stcCCFlags);
         lexer->Apply(m_stcEnv);
         lexer->Apply(m_stcCommands);
+        lexer->Apply(m_stcRemoteCommands);
     }
 
     m_dvListCtrlTargets->SetSortFunction([](clRowEntry* a, clRowEntry* b) {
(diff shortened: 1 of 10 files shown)


iHardRock:
@eranif, please check this PR :)

Plugin/FileSystemWorkspace/FSConfigPage.cpp
```diff
@@ -156,7 +164,7 @@ void FSConfigPage::Save()
         last_executables.swap(small_arr);
     }
 
-    m_config->SetExecutable(m_comboBoxExecutable->GetStringSelection());
+    m_config->SetExecutable(m_comboBoxExecutable->GetValue()); //->GetStringSelection());
```
Jarod42, line 167: I would remove commented code.
Can be retrieved from history anyway.

Plugin/FileSystemWorkspace/clFileSystemWorkspace.cpp
```diff
@@ -999,8 +1005,30 @@ void clFileSystemWorkspace::OnDebug(clDebugEvent& event)
...
+    
+    if (dbgr->GetIsRemoteDebugging()) {
+        // debugging remote target
+        wxString comm;
+        wxString host = GetConfig()->GetDebuggerRemoteHost();
+        wxString port = GetConfig()->GetDebuggerRemotePort();
+        
+        comm << host;
+
+        host = host.Trim().Trim(false);
```
Jarod42, line 1017: You probably invert those 2 lines. else modifying host is useless.

iHardRock: Yes, you right. It was blindly copied from LiteEditor/manager.cpp (around line 2238) :)
So this "bug" already present in CodeLite (https://github.com/eranif/codelite/blob/287ff78fa959f6f4491dea73ea1dc5bf92710cd3/LiteEditor/manager.cpp#L2080)

Jarod42 (approved):

eranif:
> @eranif, please check this PR :)

Sorry for the slow response time, been a busy week at work. Will look into it this weekend

PR #3553, opened 2024-12-28
Fix compilation of git.cpp in non-SFTP build.

Jarod42:


git/git.cpp
```diff
@@ -2735,6 +2731,7 @@ bool GitPlugin::DoExecuteCommandSync(const wxString& command, wxString* commandO
 
         clDEBUG() << "<-" << output.value() << endl;
         *commandOutput = output.value();
+#endif
```
Jarod42, line 2734: I assume in `!USE_SFTP`,  `IsRemoteWorkspace()` is always `false`, so no alternative code for that branch.

PR #3572, opened 2025-01-12
Improve CodeLite symlink handling

UffeJakobsen:
Subject: Improve CodeLite symlink handling

This is a new attempt at improving the symlink handling.

I know it is probably not there yet - but now I'll put it up for review :-)

I've added an attempt to have proper option handling.

Operation of FileUtils::RealPath() can be controlled in the Preferences - under the tab "Tabs"

This patch improves CodeLite with projects that have symlinks in the path

If you open a file in a project that has symlinks in its path - you'll see the expanded/resolved path of the file in the window title. Also you'll typically see that CodeLite is unable find the file in the Workspace View

There are several problems in respect to symlinks and the use of realpath() calls that this patch addresses:

 - Clangd LSP returns realpath() paths

 - mainbook uses realpath() extensively to figure out if a file is already opened (in another tab).

 - CodeLite Find-In-Files uses realpath() extensively (and in the end - clicking on the found file opens it)

 - etc...

This patch basically fakes the use of FileUtils::RealPath() calls - on most occasions just returning the unmodified path.

An optional argument to FileUtils::RealPath() (defaults to false) can force the realpath() conversion (this is only used in very few occasions in this patch)


LiteEditor/editorsettingsdockingwidows.cpp
eranif, line 1: This change should be under the "Misc" section, "Tabs" is more for the visualization of the tab control (height, orientation etc)

Plugin/optionsconfig.cpp
eranif, line 1: I am trying to avoid adding new entries to the `OptionsConfig`, please use `clConfig` instead where it makes sense. i.e. if the configuration entry is part of a "family" of entries that are already exist in `OptionsConfig` then it should go there, otherwise, we should use `clConfig`

CodeLite/fileutils.cpp
```diff
@@ -617,6 +620,17 @@ wxString FileUtils::RealPath(const wxString& filepath)
     return filepath;
 }
 
+bool FileUtils::RealPathGetModeResolveSymlinks(void)
```
eranif, line 623: Was this code formatted using `clang-format` ? I suspect not

eranif (changes requested):

Plugin/optionsconfig.cpp
UffeJakobsen: Great - that makes much easier - also nice to avoid more xml-stuff :-)

CodeLite/fileutils.cpp
```diff
@@ -617,6 +623,10 @@ wxString FileUtils::RealPath(const wxString& filepath)
     return filepath;
 }
 
+bool FileUtils::RealPathGetModeResolveSymlinks(void) { return bRealPathModeResolveSymlinks; }
```
Jarod42, line 626:
```suggestion
bool FileUtils::RealPathGetModeResolveSymlinks() { return bRealPathModeResolveSymlinks; }
```

CodeLite/fileutils.h
```diff
@@ -247,7 +247,9 @@ class WXDLLIMPEXP_CL FileUtils
     /**
      * @brief (on Linux) makes-absolute filepath, and dereferences it and any symlinked dirs in the path
      */
-    static wxString RealPath(const wxString& filepath);
+    static wxString RealPath(const wxString& filepath, bool forced=false);
+    static bool RealPathGetModeResolveSymlinks(void);
```
Jarod42, line 251:
```suggestion
    static bool RealPathGetModeResolveSymlinks();
```

CodeLite/fileutils.cpp
```diff
@@ -62,6 +62,8 @@
 #include <memory>
 #include <wx/filename.h>
 
+static bool bRealPathModeResolveSymlinks = true;
```
Jarod42, line 65: Would it make sense to promote the static class in non static one and have it as regular member?

UffeJakobsen: I'm ok with the current solution - but in the end it is up to Eranif is guess

eranif (approved):

PR #3578, opened 2025-01-20
Remember Find/Replace dialog position and ideal size

UffeJakobsen:
This change makes Find/Replace dialog remember its position between restarts.
Also is rearranges the dialog to its ideal height and width - where all buttons and fields are visible

### DIFF
--- a/LiteEditor/quickfindbar.cpp
+++ b/LiteEditor/quickfindbar.cpp
@@ -250,13 +250,12 @@ QuickFindBar::QuickFindBar(wxWindow* parent, wxWindowID id)
     if (clConfig::Get().Read("FindBar/Width", dlg_width) && clConfig::Get().Read("FindBar/Height", dlg_height)) {
         SetSize(dlg_width, dlg_height);
         SetSizeHints(dlg_width, dlg_height);
-        SetMinClientSize(min_size);
     } else {
         // first time, place it at the top
         Move(wxNOT_FOUND, parent->GetPosition().y);
-        SetMinClientSize(min_size);
-        GetSizer()->Fit(this);
     }
+    SetMinClientSize(min_size);
+    GetSizer()->Fit(this);
     Layout();
 }
 


UffeJakobsen:
Please note that since commit https://github.com/eranif/codelite/commit/daf77af07614287757acdb2b361ab0b4921df48b - this issue have returned

Seen @ Archlinux / stable wx-3.2.8

eranif:
I just tested this and the find & replace size and position is persistent between restarts. Isn't this the case for you?

UffeJakobsen:
> I just tested this and the find & replace size and position is persistent between restarts. Isn't this the case for you?

No not really.

After restart the Find/Replace dialog opens up in (0,0)
The at first open it does get focus either.
If I move and resize the Find/Replace dialog - it is (only) remembered for that Codelite session
after restart of CL it is back to 0,0 - but with its old size remembered

(Archlinux/wx-3.2.8.1)

eranif:
I suspect a wx issue, I don't have wx3.2 but it is working as expected on wx-master. I will try and test it with wx3.2

UffeJakobsen:
> I suspect a wx issue, I don't have wx3.2 but it is working as expected on wx-master. I will try and test it with wx3.2

I just tried to build wx-.3.3.x - but my first attempt failed... will give it another try later

UffeJakobsen:
I did a build with wx-3.3 - it has the same problem as described above
What platform are you running/testing on ?

eranif:
I tried this on Windows & macOS. I will build on Linux to double check (WSL2/Ubuntu24.04)

eranif:
See this animated gif, works on Ubuntu 24.04.
https://codelite.org/tmp/f_n_r.gif

UffeJakobsen:
Hmm strange - I'll have to try another distro - but that may have to wait until some time next week.
I'll update this issue with my discoveries :-)

PR #3605, opened 2025-03-17
Adding missing fields

bztd:


Runtime/org.codelite.codelite.metainfo.xml
```diff
@@ -23,4 +28,16 @@
...
+  <provides>
+    <binary>codelite</binary>
+  </provides>
+
+  <content_rating type="oars-1.1">
+    <content_attribute id="social-info">mild</content_attribute>
+  </content_rating>
+  <releases>
+	<release version="18.1.0" date="2025-01-16"></release>
+  </releases>
```
Jarod42, line 41: Is that part required?
As it would require to update that file too for each release.

bztd: Yes, it's necessary to update with each version. This file is used by services like Flatpak to obtain that information.

Jarod42 (approved):

PR #3608, opened 2025-03-19
Statusbar: ensure that statusbar elements can be controlled via preferences - make field properly sized for its contents

UffeJakobsen:
It gets rather lengthy...

```
-    int lineColWidth = GetTextWidth("Ln 100000, Col 999, Pos 12345678, Len 4821182");
+    int lineColWidth = GetTextWidth("Ln 123456, Col 123, Pos 12345678, Len 12345678, Sel 1234567, SelLn 12345");
```

Another approach could be to not display "Pos" and "Len" - if a selection is ongoing ?


Jarod42 (approved):

eranif (commented):
What to show is controllable from `settings -> preferences -> misc -> status bar`
You can make new entry for the new addition and let the user decide what he wants to display

Jarod42 (approved):

eranif (approved):

PR #3618, opened 2025-04-01
[submodule] Bump dtl version from custom eranif fix to official v1.21 (13th May 2024)

Jarod42:


Jarod42:
msys CI now fails because of an update of CMake to CMake 4.
Which is unrelated to that PR.

eranif (approved):

PR #3619, opened 2025-04-01
Fix #3617 (Incorrect use of `std::unique_ptr<char>`) mixing `new[]` with `delete`

### LINKED ISSUE #3617: Incorrect use of std::unique_ptr<char> instead of std::unique_ptr<char[]> for array allocation
In CodeLite/winprocess.cpp, there's a potential memory deallocation bug due to the incorrect use of std::unique_ptr<char> to manage a dynamically allocated character array.

Problematic code (line 180 as of current commit):

```
char* chBuf = new char[65536 + 1]; // 64K should be sufficient buffer
memset(chBuf, 0, 65536 + 1);

std::unique_ptr<char> sp(chBuf); // Incorrect - this calls delete instead of `delete[]`
```
This will result in undefined behavior when the unique_ptr goes out of scope, since it calls delete on an array allocated with new[].

Suggested fix:

Replace with std::unique_ptr<char[]>

Jarod42:
closes #3617

### DIFF
--- a/CodeLite/winprocess.cpp
+++ b/CodeLite/winprocess.cpp
@@ -173,26 +173,22 @@ WinProcess::~WinProcess() {}
 
 bool WinProcess::Read(wxString& buff)
 {
-    DWORD dwRead;
-    DWORD dwMode;
-    DWORD dwTimeout;
-    char* chBuf = new char[65536 + 1]; // 64K should be sufficient buffer
-    memset(chBuf, 0, 65536 + 1);
-
-    std::unique_ptr<char> sp(chBuf);
+    constexpr auto bufferSize = 65536; // 64K should be sufficient buffer
+    std::vector<char> chBuf(bufferSize + 1, '\0'); 
 
     // Make the pipe to non-blocking mode
-    dwMode = PIPE_READMODE_BYTE | PIPE_NOWAIT;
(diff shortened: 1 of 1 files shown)


Jarod42:
msys2 now fails because of new CMake 4 on msys2. Unrelated to that PR

CodeLite/winprocess.cpp
```diff
@@ -173,26 +173,22 @@ WinProcess::~WinProcess() {}
 
 bool WinProcess::Read(wxString& buff)
 {
-    DWORD dwRead;
-    DWORD dwMode;
-    DWORD dwTimeout;
-    char* chBuf = new char[65536 + 1]; // 64K should be sufficient buffer
-    memset(chBuf, 0, 65536 + 1);
-
-    std::unique_ptr<char> sp(chBuf);
```
eranif, line 182: wouldn't be sufficient just to change this into: 

```c++
std::unique_ptr<char[]> sp(chBuf);
```

eranif (approved):

eranif (approved):

CodeLite/winprocess.cpp
Jarod42: It was possible too. `std::vector` seems simpler/more appropriate IMO.

PR #3634, opened 2025-04-25
[CI][MacOS] Use `-DCMAKE_POLICY_VERSION_MINIMUM=3.5` to bypass compatibility error

Jarod42:


Jarod42:
I don't have MacOS.
error from CI

> -- Configuring incomplete, errors occurred!
CMake Error at cmake/Modules/OSXDeps.cmake:5 (execute_process):
  execute_process failed command indexes:
>
> -- Configuring incomplete, errors occurred!
    1: "Child return code: 1"
>
> Call Stack (most recent call first):
  cmake/Modules/OSXDeps.cmake:34 (build_and_install_submodule)
  CMakeLists.txt:45 (include)


Not sure if we should add `-DCMAKE_POLICY_VERSION_MINIMUM=3.5` in the `execute_process`... testing

PR #3642, opened 2025-05-12
update chinese

tuwulin365:
update chinese


Jarod42 (approved):
Approve mostly blindly: *"it is Chinese for me"* ;-)

PR #3650, opened 2025-06-01
[cleanup] Move clMainFrameHelper files from Plugin to LiteEditor

Jarod42:


eranif (changes requested):
The whole purpose of this file was to expose some of clMainFrame API into the plugin, this is why it is located under Plugin folder. Code that is placed inside the "Plugin" folder or "CodeLite" folder can not access classes inside from the "LiteEditor" folder.

If we can remove this class into LiteEditor, I guess that we can remove it completely...

Jarod42:
> Code that is placed inside the "Plugin" folder or "CodeLite" folder can not access classes inside from the "LiteEditor" folder.

I understand that,
so the existing comment with danger

> If we can move this class into LiteEditor, I guess that we can remove it completely...

Is there some non-official Codelite plugins known?
I will remove the class completely then.

eranif (approved):

PR #3675, opened 2025-07-17
Update windows template projects

Jarod42:
+ Fix copy-paste "typo" Linux->Windows

`Load` currently tries to `Save` outdated projects, but installDirectory (C:/Programs) need admin right.
As the `Save` fails (because of permission), the `Load` is considered as failing too, so the the project is not available (and log show an error).

Updating the project avoids those issues.


Jarod42:
> msys2 / msys2 (master) (pull_request) Failing

Unrelated to that change.

wxWidget master seems to have changed output directory/name of one of their dll, making our CI fails.
 I will investigate how to fix it.

PR #3737, opened 2025-10-31
Change more of the installation file layout to match Linux under Posix layout

stahta01:


wxcrafter/src/wxgui_bitmaploader.cpp
```diff
@@ -14,7 +14,12 @@ wxCrafter::ResourceLoader::ResourceLoader(const wxString& skin)
 {
     if(m_bitmaps.empty()) {
         wxString zipFile;
+#ifdef USE_POSIX_LAYOUT
+        zipFile << wxStandardPaths::Get().GetDataDir() + wxT(INSTALL_DIR) << wxFileName::GetPathSeparator() << skin
```
Jarod42: Strange to mix `+` and `<<`

Jarod42 (approved):

wxcrafter/src/wxgui_bitmaploader.cpp
stahta01: I can change the + to << if you think that is more correct or common.
I will test it and verify it works.

stahta01: I did the change and the [ERR] did not return so the change worked as well as the prior code to fix the issue.

eranif:
How much effort will it take to convert this change to be the only installation on Windows?
Today, when you run `make install` under window, it installs everything under `${CMAKE_BINARY_DIR}/install` and the setup installer simply packages this folder. I don't mind having a single installation layout on Windows and using this one as it.

stahta01:
> How much effort will it take to convert this change to be the only installation on Windows? Today, when you run `make install` under window, it installs everything under `${CMAKE_BINARY_DIR}/install` and the setup installer simply packages this folder. I don't mind having a single installation layout on Windows and using this one as it.

It should already be that way. Since nearly everything is in a if/else statement. Edit: It should only affect the MINGW install layouts.
Edit: Reread your post you want me to get rid of the other MINGW installation layout, is that correct? Because I would think that is a little soon for that to be done.

stahta01:
> > How much effort will it take to convert this change to be the only installation on Windows? Today, when you run `make install` under window, it installs everything under `${CMAKE_BINARY_DIR}/install` and the setup installer simply packages this folder. I don't mind having a single installation layout on Windows and using this one as it.
> 
> It should already be that way. Since nearly everything is in a if/else statement. Edit: It should only affect the MINGW install layouts. Edit: Reread your post you want me to get rid of the other MINGW installation layout, is that correct? Because I would think that is a little soon for that to be done.

But if you think it is a good idea, I will start the change to always use the Posix layout under windows.

Tim S.

PR #3742, opened 2025-11-04
Change Windows install to always use the Posix layout

stahta01:
The last PR I did had a request that I change to always doing Posix layout for windows. The third commit does that.
The first two fixes the windows installer by adding the missing DLL that resulted in a run-time error. I think it is a little early to switch to the Posix layout; but maintaining two layouts for windows is likely just as much trouble as finding the minor issues of switching at this time. I do have concerns about the iss code that removes the installation I really do not understand it enough to know how to update it. It has been 20+ years since I did Inno Setup and I never worried about the uninstall code at that time.

Tim S.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,8 +35,7 @@ cmake_minimum_required(VERSION 3.16)
 # Windows / MSYS options:
 #
 # -DWXWIN=<fullpath>                          //
-# https://docs.codelite.org/build/build_wx_widgets/#windows -DWITH_POSIX_LAYOUT
-# // When building under Windows, install CodeLite using POSIX layout
+# https://docs.codelite.org/build/build_wx_widgets/#windows
 #
 # ##############################################################################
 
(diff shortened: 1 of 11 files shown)


Jarod42 (approved):

eranif:
Let me apply this PR locally to check the upgrade path before we merge it

Jarod42:
> Let me apply this PR locally to check the upgrade path before we merge it

Didn't test it, but there are installers as artefact available at https://github.com/eranif/codelite/actions/runs/19054679733

And there is a "Show content of installation" too as step which looks good to me.

eranif:
@stahta01 FYI: 

I am getting this error during `make -j$(nproc) install`

```bash
file INSTALL cannot find "C:/msys64/clang64/bin/libuchardet.dll": File
```

I assume that it is needed because of `UCRT64` ?

eranif:
For now, I changed it to include this DLL on UCRT64 systems only, let me know if this logic is correct.

stahta01:
I did my testing on CLANG64 and it should have worked for you; you did apply all 3 commits because the first one installed the package with the DLL.

eranif:
I don't have this DLL on my system. However, on this commit: 
https://github.com/eranif/codelite/pull/3742/commits/05fa8925bbebb4ac4f38b131aa5677e9ccb0dc2e you added new library which was never required before this PR (`mingw-w64-clang-x86_64-uchardet`).

stahta01:
> I don't have this DLL on my system. However, on this commit: [05fa892](https://github.com/eranif/codelite/commit/05fa8925bbebb4ac4f38b131aa5677e9ccb0dc2e) you added new library which was never required before this PR (`mingw-w64-clang-x86_64-uchardet`).

Are you using Windows 10 or 11 to test? I am using Win11 and codelite refused to run without that DLL.

Tim S.

eranif:
Windows 11. I think that your build system is "contaminated". All the releases that I build for Windows never had this DLL. You can download an installer from the website and see that this DLL is not included and yet CodeLite runs fine.

stahta01:
> Windows 11. I think that your build system is "contaminated". All the releases that I build for Windows never had this DLL. You can download an installer from the website and see that this DLL is not included and yet CodeLite runs fine.

Since it is less than two weeks old not likely; but it is still worth a try to confirm.

Edit: Maybe I had the issue before I re-installed MSys2.

Tim S.

eranif:
> Since it is less than two weeks old not likely; but it is still worth a try to confirm.

By "contaminated" I meant: you probably installed different toolchain than the one I recommend on the website - which "pulled" in a new DLL, that's all. That would be my guess...

PR #3744, opened 2025-11-05
CMake: Change CMAKE_INSTALL_PREFIX cache type to PATH

carlo-bramini:
Configuring codelite with CMake prints this message on the console:
```
CMake Warning (dev) at CMakeLists.txt:71 (set):
  implicitly converting 'ON' to 'STRING' type.
This warning is for project developers.  Use -Wno-dev to suppress it.
```
This warning looks correct to me.
The right type to apply to that `set()` statement should be `PATH`, in my opinion.


Jarod42:
Note: `-Wno-dev` is used for both msys2 and MacOS CIs

PR #3746, opened 2025-11-07
Remove uchardet code added by mistake

stahta01:
I replaced the MSys2 installation as suggested by you and also did "git clean -fxd" and the uchardet DLL was no longer needed to run codelite after a rebuild of it. Edit: I removed the suggested line and also the line before and after.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1192,9 +1192,6 @@ if(NOT WXC_APP)
 endif()
 
 if(MINGW)
-  set(MSYSTEM_ENV_VAR "$ENV{MSYSTEM}")
-  message(STATUS "MSYSTEM_ENV_VAR => ${MSYSTEM_ENV_VAR}")
-
   # determine MSYS2 path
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -1216,11 +1216,6 @@ if(MINGW)
     libhunspell-1.7-0.dll
     libunwind.dll)
 
-  if("${MSYSTEM_ENV_VAR}" STREQUAL "UCRT64")
-    message(STATUS "Running on UCRT64, including libuchardet.dll DLL")
-    list(APPEND SYSTEM_DLLS libuchardet.dll)
-  endif()
```
eranif, line 1222: We should also remove line `1196` as well.

stahta01: I will look at line 1196

stahta01: I removed 1196 and the line before and after

stahta01:
I removed the suggested line and also the line before and after.

eranif (approved):

PR #3747, opened 2025-11-08
[cleanup] Migrate `clCxxWorkspace::CreateCompileCommandsJSON()` to `nlohmann::json`

Jarod42:
And split function

### DIFF
--- a/Plugin/workspace.h
+++ b/Plugin/workspace.h
@@ -26,7 +26,6 @@
 #define WORKSPACE_H
 
 #include "IWorkspace.h"
-#include "JSON.h"
 #include "clWorkspaceManager.h"
 #include "cl_command_event.h"
 #include "codelite_exports.h"
@@ -37,6 +36,7 @@
 #include "singleton.h"
 #include "wxStringHash.h"
 
+#include <assistant/common/json.hpp> // <nlohmann/json.hpp>
 #include <map>
 #include <wx/event.h>
 #include <wx/filename.h>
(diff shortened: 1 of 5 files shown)


Plugin/workspace.h
```diff
@@ -174,10 +174,15 @@ class WXDLLIMPEXP_SDK clCxxWorkspace : public LocalWorkspaceCommon
     wxArrayString GetWorkspaceFolders() const;
 
     /**
-     * @brief create 'compile_commands' json object for the workspace projects (only the enabled ones)
-     * or compile_flags.txt file
+     * @brief create 'compile_commands' json object for the workspace (enabled) projects
      */
-    cJSON* CreateCompileCommandsJSON(bool createCompileFlagsTxt, wxArrayString* generated_paths) const;
+    nlohmann::json CreateCompileCommandsJSON() const;
```
eranif, line 179: Let's add a new header somewhere (in the `ROOT/CodeLite/` folder) with the following content:

```c++
#pragmace once

#include <assistant/common/json.hpp>

using clJSON = nlohmann::json;
```

I have already "burnt" once when I exposed the JSON implementation (`cJSON`) in various places. This will help any future migrations.

WDYT?

Jarod42: The `using ClJSON` seems not really help migration, as json API might change, and it seems overkill/complicated to wrap json library to have "stable" interface.
But it can be done (it is Plugin which depends of assistant, not libcodelite, so putting it in Codelite seems not fully correct neither).

In addition, I have to investigate the "correct" way to handle shared dependencies, as using nlohmann from assistant seems hacky (it was already in my TODO list for (wx)sqlite3 dependency).

PR #3752, opened 2025-11-13
Clear and to switch to output tab when launching program without debugging

mvohla:
Add options to clear and to switch to output tab when launching a program without debugging. 
Adds a section "Launch" in Preferences -> Settings -> Misc.
This is just a convenience feature that suits my personal workflow.
Translations for the options have not been added.

### DIFF
--- a/LiteEditor/manager.cpp
+++ b/LiteEditor/manager.cpp
@@ -1688,14 +1688,14 @@ void Manager::ExecuteNoDebug(const wxString& projectName)
 
     m_programProcess = ::CreateAsyncProcess(this, execLine, createProcessFlags, wd, &env_list);
     if (m_programProcess) {
-        if(clConfig::Get().Read(kConfigClearOutputOnLaunch, false)) {
+        if (clConfig::Get().Read(kConfigClearOutputOnLaunch, false)) {
             clGetManager()->ClearOutputTab(kOutputTab_Output);
         }
         clGetManager()->AppendOutputTabText(
             kOutputTab_Output, wxString() << _("Working directory is set to: ") << wd << "\n", false);
         clGetManager()->AppendOutputTabText(
             kOutputTab_Output, wxString() << _("Executing: ") << execLine << "\n", false);
-        if(clConfig::Get().Read(kConfigShowOutputOnLaunch, false)) {
+        if (clConfig::Get().Read(kConfigShowOutputOnLaunch, false)) {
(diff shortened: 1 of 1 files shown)


LiteEditor/manager.cpp
```diff
@@ -1688,11 +1688,16 @@ void Manager::ExecuteNoDebug(const wxString& projectName)
 
     m_programProcess = ::CreateAsyncProcess(this, execLine, createProcessFlags, wd, &env_list);
     if (m_programProcess) {
+        if(clConfig::Get().Read(kConfigClearOutputOnLaunch, false)) {
```
Jarod42: Wrong formatting

mvohla: Thank you for reviewing! Whitespace after "if" was added. Anything else?

Jarod42: Not from my side.

Eric-01:
If I may put my two cents in, the word "launch" used in the new settings header and checkbox labels feels a bit too vague in my opinion, which isn't necessarily a bad thing by itself, but since there's already another section nearby called "Startup", I see a lot of potential for users to confuse one with the other.

It also doesn't help that there has been no support for hover tooltips for individual options since the preferences window rewrite in CodeLite 16.1.0 in commit abc5d0e0bb099c1affaef87fdcba0d207d98f433 and the subsequent ones (which I was never a fan of, and I still feel the same way today). Because of that, you can no longer simply keep the checkbox label short and move all the details into a tooltip so that everyone is happy.

Instead, you have to make compromises, which leads to some options — like the new one from the latest release — ending up with labels such as: "Find next/Find previous actions override search string with current selection (like Find next at caret/prev at caret)". This happened because there are no tooltips available since CodeLite 16.1.0 to move the additional explanation to, and the author apparently still wanted to make sure everyone understands what the option does, so we ended up with this monstrosity.

Maybe it would be possible to clarify that this newly added section, along with its options, refers specifically to running the project you're currently working on (`Ctrl + F5` / `Ctrl + F9`), rather than the entire IDE (which is what the aforementioned "Startup" section is for). The question is, how can this be done? Maybe a change in terminology or something similar would help here...

eranif:
> It also doesn't help that there has been no support for hover tooltips for individual options since the preferences window rewrite in CodeLite 16.1.0 in commit

Point taken (in a good way), I will see what I can do about this

PR #3769, opened 2025-11-26
[cleanup] Migrate `CodeFormatter` to `nlohmann::json`

Jarod42:


CodeFormatter/CodeFormatterManager.cpp
```diff
@@ -71,18 +73,22 @@ void CodeFormatterManager::Load()
         return;
     }
 
-    JSON root{ config_file };
-    if (!root.isOk() || !root.toElement().isArray()) {
+    wxString content;
+    if (!FileUtils::ReadFileContent(config_file, content))
```
eranif, line 77: Please run `clang-format`

CodeFormatter/GenericFormatter.cpp
```diff
@@ -204,18 +208,18 @@ bool GenericFormatter::FormatString(const wxString& content, const wxString& ful
     return true;
 }
 
-void GenericFormatter::FromJSON(const JSONItem& json)
+void GenericFormatter::FromJSON(const nlohmann::json& json)
 {
     SourceFormatterBase::FromJSON(json);
-    m_command = json["command"].toArrayString();
-    m_workingDirectory = json["working_directory"].toString();
+    m_command = JsonUtils::ToArrayString(json["command"]);
```
eranif, line 217: This might throw an exception, lets wrap it with `try` / `catch` block.
I suggest that we add new methods in `JsonUtils`.

For example:

```c++
template <typename T>
concept JsonExtractable = requires(const nlohmann::json& j)
{
    {j.template get<T>()}->std::same_as<T>;
};

template <typename T>
inline std::optional<T> GetValue(const nlohmann::json& json, const std::string& name) requires JsonExtractable<T>
{
    if (!json.contains(name)) {
        return std::nullopt;
    }
    try {
        return json[name].get<T>();
    } catch (const std::exception&) {
        return std::nullopt;
    }
}
```

We can later use with default value like this:

```c++
m_command = wxString::FromUTF8(JsonUtils::GetValue(json, "command").value_or(""));
```

CodeFormatter/SourceFormatterBase.cpp
```diff
@@ -1,31 +1,31 @@
...
-void SourceFormatterBase::FromJSON(const JSONItem& json)
+void SourceFormatterBase::FromJSON(const nlohmann::json& json)
 {
-    m_languages = json["languages"].toArrayString();
-    m_flags = json["flags"].toSize_t((size_t)FormatterFlags::ENABLED);
-    m_configFile = json["config_file"].toString();
-    m_name = json["name"].toString();
-    m_description = json["description"].toString();
-    m_shortDescription = json["short_description"].toString();
+    m_languages = JsonUtils::ToArrayString(json["languages"]);
```
eranif, line 15: See previous comment about exceptions in  `nlohmann::json`

eranif (changes requested):

CodeFormatter/GenericFormatter.cpp
Jarod42: I noticed afterward that all methods of `JsonItem` defaults it return value when "empty", whereas nlohmann throws exception for invalid access.
`nlohmann::json::value(key, default_value)` exists for that case too.

I have indeed to reworks my migrations.

PR #3770, opened 2025-11-26
[cleanup] Migrate wxcrafter to `nlohmann::json`

Jarod42:


wxcrafter/src/wxc_settings.cpp
```diff
@@ -20,45 +23,48 @@ wxcSettings& wxcSettings::Get()
...
-    root.toElement().append(arr);
+    auto& arr = root["m_templateClasses"];
 
-    for (const auto& p : m_templateClasses) {
-        arr.append(p.second.ToJSON());
+    for (const auto& [_, customControlTemplate] : m_templateClasses) {
+        arr.push_back(customControlTemplate.ToJSON());
     }
-    root.save(fn);
+    FileUtils::WriteFileContentRaw(fn, root.dump());
```
eranif, line 67: do we have default for `dump`?, You used `2` everywhere else

Jarod42: 0 is default (for no spacing).
Not sure which one you prefer
For inspecting json, formatting can be done after in editor.

Jarod42:
I will do the change differently to remove cJSON, modifying JSON.h instead (at a first step at least)

PR #3834, opened 2026-02-08
[UTs] Add some UTs for Codelite Json

Jarod42:


Tests/CodeliteTest/JsonTests.cpp
```diff
@@ -0,0 +1,296 @@
...
+    CHECK(item.arraySize() == 3);
+    CHECK(item[0].isNumber());
+    CHECK(item[1].isBool());
+    CHECK(item[2].isNull());
+
+    CHECK(!item.hasNamedObject("a"));
+    CHECK(!item.namedObject("a").isOk());
+    CHECK(!item["a"].isOk());
+#if 0 // crash
+    CHECK(item.GetAsMap().empty());
```
eranif, line 221: Are we okay with this?

Jarod42: It should probably be fixed indeed.
Purpose of that PR is to add test to ease refacto and fixes.

PR #3835, opened 2026-02-11
Replace wxsf-code Submodule With Inline wxxmlserializer Sources

eranif:
Remove the wxsf-code git submodule and embed the required wxxmlserializer library sources directly into the DatabaseExplorer plugin. This eliminates the external dependency on wxshapeframework while retaining only the XML serialization functionality that DatabaseExplorer actually uses.

The change includes adding header files (XmlSerializer.h, PropertyIO.h, Defs.h) and implementation files (XmlSerializer.cpp, PropertyIO.cpp) directly under DatabaseExplorer/wx/wxxmlserializer/. The CMakeLists.txt is updated to remove the wxshapeframework library dependency and add the new source directory to include paths.

* DatabaseExplorer: Add inline wxxmlserializer sources
* submodules: Remove wxsf-code submodule reference
* CMakeLists.txt: Remove wxshapeframework build and dependencies

** Generated by CodeLite. **


DatabaseExplorer/wx/wxxmlserializer/Defs.h
```diff
@@ -0,0 +1,7 @@
+#ifndef _XSDEFS_H
+#define _XSDEFS_H
+
+#define WXDLLIMPEXP_XS
+#define WXDLLIMPEXP_DATA_XS(type) type
```
Jarod42, line 5: I will probably remove that stuff afterward as cleanup task :-)

Jarod42 (approved):

PR #3840, opened 2026-02-15
Fix build for GCC

stahta01:
I build under MSys2 MinGW UCRT64 with other patches needed to build; but this patch is needed for GCC to fix an error caused by missing include.

Tim S.


eranif:
Thanks, I rebased and merged it. The failing CI was already fixed in master branch.

PR #3854, opened 2026-02-28
[cleanup] Remove unneeded `StdToWX::ToArrayString` calls

Jarod42:


eranif:
Is this supported in wx3.2 on all build types or does it require an STL build of wxWidgets? (which in this case,  `wxArrayString` is a subclass of `class wxBaseArray : public wxVector<T> `? (where `wxVector` is `using wxVector = std::vector<T>;`)

Jarod42:
Not sure, I will add a CI build with `wxUseSTL=0` (or `--disable-stl`) to check.

Jarod42:
Added only to macOS CI (`--disable-stl`) and it passes :-)

PR #3882, opened 2026-03-19
Dropping `ctagsd` LSP.

eranif:


CMakeLists.txt
```diff
@@ -1113,9 +1113,7 @@ else()
 
   # utility for importing themes
   add_subdirectory(codelite-generate-themes)
-
   add_subdirectory(codelite_echo)
-  add_subdirectory(ctagsd)
```
Jarod42, line 1118: That directory is not removed.
Is it normal?

eranif: You are right, I thought of keeping it (maybe I am getting too sentimental 🤣 )
I will removing it.

PR #3896, opened 2026-03-23
Replace `cJSON` by `nlohmann::ordered_json`

Jarod42:
Using ordered version to better mimic old behavior which keeps insertion order


Jarod42:
@eranif All is ok on my side.
I missed parsing-handling in the first version (we try to parse (xml) workspace as json ?!).
You might do further testing if you want :-)

eranif:
I am just going to merge this and test it, hopefully nothing will break 🤣

eranif:
@Jarod42 so far so good...

PR #3902, opened 2026-03-30
Json nlohmann

Jarod42:
Use `nlohmann::json` in most places where `JSONItem::createObject()` were used.


DebugAdapterClient/clDapSettingsStore.cpp
```diff
@@ -4,15 +4,13 @@
...
-    json.addProperty("flags", m_flags);
-    json.addProperty("env_format", (int)m_envFormat);
-    json.addProperty("launch_type", (int)m_launch_type);
-    return json;
+    return nlohmann::json{{"name", StringUtils::ToStdString(m_name)},
+                          {"command", StringUtils::ToStdString(m_command)},
+                          {"connection_string", StringUtils::ToStdString(m_connection_string)},
+                          {"environment", StringUtils::ToStdString(m_environment)},
+                          {"flags", m_flags},
+                          {"env_format", m_envFormat},
```
eranif, line 12: Do we need to preserve the conversion here? old code used `(int)`

CodeLite/LSP/basic_types.cpp
```diff
@@ -41,9 +41,7 @@ void TextDocumentIdentifier::FromJSON(const JSONItem& json)
 
 JSONItem TextDocumentIdentifier::ToJSON() const
 {
-    JSONItem json = JSONItem::createObject();
-    json.addProperty("uri", GetPathAsURI());
-    return json;
+    return nlohmann::json{"uri", StringUtils::ToStdString(GetPathAsURI())};
```
eranif: do we miss a `{` here?

I thought it should be:

```c++
return nlohmann::json{{"uri", StringUtils::ToStdString(GetPathAsURI())}};
```

abbreviation/abbreviationentry.cpp
```diff
@@ -48,9 +48,4 @@ void AbbreviationJSONEntry::FromJSON(const JSONItem& json)
 }
 
 JSONItem AbbreviationJSONEntry::ToJSON() const
-{
-    JSONItem ele = JSONItem::createObject();
-    ele.addProperty("entries", m_entries);
-    ele.addProperty("autoInsert", m_autoInsert);
-    return ele;
-}
+{ return nlohmann::json{{"entries", JsonUtils::ToJson(m_entries)}, {"autoInsert", m_autoInsert}}; }
```
eranif, line 51: I thought that we are using `nlohmann::ordered_json` why `nlohmann::json` now, is this intentionally?

CodeFormatter/SourceFormatterBase.cpp
```diff
@@ -17,14 +17,12 @@ void SourceFormatterBase::FromJSON(const JSONItem& json)
...
-    ele.addProperty("languages", m_languages);
-    ele.addProperty("flags", m_flags);
-    ele.addProperty("config_file", m_configFile);
-    ele.addProperty("name", m_name);
-    ele.addProperty("description", m_description);
-    ele.addProperty("short_description", m_shortDescription);
-    return ele;
+    return nlohmann::json{{"languages", JsonUtils::ToJson(m_languages)},
+                          {"flags", m_flags},
+                          {"config_file", StringUtils::ToStdString(m_configFile)},
```
eranif: I think that since we are using UTF-8 everywhere, we can simply used: `m_configFile.ToStdString(wxConvUTF8)`

DebugAdapterClient/clDapSettingsStore.cpp
Jarod42: Conversion is done implicitly from json side, enum would be translated into `int` (the underlying type).
Explicit conversion is unneeded. But I can revert back if you prefer explicit one.

CodeLite/LSP/basic_types.cpp
Jarod42: Indeed.

abbreviation/abbreviationentry.cpp
Jarod42: I mostly use `ordered_json` in JSonItem so that in tests, we might have predictive expectation with `format`.

One is convertible into the other.

I think `json` is most commonly used than `ordered_json`. I might use the one you prefer.

CodeFormatter/SourceFormatterBase.cpp
Jarod42: So we might remove `StringUtils::ToStdString` everywhere (not only in json)?

I would be provide specialization for `adl_serializer<wxString>` to avoid to do that everywhere in json, but it requires to disable some implicit conversions from nlohmann::json.
There is a PR in that direction in assistant for that.

eranif (approved):

PR #3934, opened 2026-05-04
[CMake][wxcrafter] Allow to generate files from *.wxcp from commandline

Jarod42:
Allow to generate files from *.wxcp from command line and use it in the build.

It is working OK on MacOS :-)
~but I got dll issues on mingw :-/~,
~and GTK issue on ubuntu CI~ (probably because of no gui) :-/

If you have insights/advices for fixes, then I would happy.


eranif:
I will pull this one to my mac and give it a go

Jarod42:
I got workaround for ubuntu.
For mingw, api-ms-win-crt-environment-l1-1-0.dll is missing :-/

Jarod42:
Issue on ubuntu/msys2 fixed

eranif:
So if I understand this correctly, you added a new macro: `wxcrafter_target` which basically generates files from `wxcp` files and these files are collected with the "GLOB" command?

I also saw that you moved `wxcrafter` to be first in build so it will be available for plugins.
However, this will not work for wxcp files defined in "Plugin" or "CodeLite".

In general this will allow us to clean our repository from generated files which also fit to other "auto-generates" tools integration with CMake, like "protobuf".

eranif (approved):

PR #3937, opened 2026-05-11
Restore backup (for copyright plugin)

Jarod42:
[cleanup] Remove `WriteFileWithBackup` which ignore backup.

If backup of copyright should be removed instead, check box option should be removed too.

### DIFF
--- a/CodeLite/fileutils.h
+++ b/CodeLite/fileutils.h
@@ -62,7 +62,7 @@ public:
 
 public:
 
-    static bool Backup(const wxString& file_name);
+    static std::optional<wxFileName> Backup(const wxString& file_name);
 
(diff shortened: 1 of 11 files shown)


CodeLite/fileutils.h
```diff
@@ -61,6 +61,9 @@ class WXDLLIMPEXP_CL FileUtils
     };
 
 public:
+
+    static bool Backup(const wxString& file_name);
```
eranif: Maybe we should change the API to return `std::optional<wxFileName>` with the backup file path?

Jarod42: Will do.

Jarod42: Done.

PR #3952, opened 2026-05-30
Don't modify wxcp when generation is done from command line

Jarod42:


eranif:
Another issue that I noticed:
I ran `cmake` on my macOS and I noticed that the `wxcp` file is modified.

It added this line:

```json
"m_generatedFilesDir": "/Users/eran/devl/codelite/.build-release/Tail",
```
I see 2 problems with this:
1. Absolute path - fixable
2. The build directory is problematic, even if we keep this relative, it will still be changed when we switch from release -> debug build.
3. Switching between Windows / Linux / macOS will cause modifications (path separator `/` and `\`) - fixable

To fix `1`: lets use relative path
To fix `2`: ??
To fix `3`: always use forward slash `/` (even on windows)

eranif (approved):

Jarod42:
I don't have MacOS, but with that PRs, I no longer modify wxcp locally.

I agree that paths should be saved preferably as relative with `/`.
I didn't change saving logic.
I just provided value from command line.
That PR should stop to save wxcp when generating from command line.

I think that `"m_generatedFilesDir"` should not be part of project (as some other fields).
From inside wxCrafter app/plugin, I think 'generate' should ask for directory

PR #3992, opened 2026-06-09
Remove generated files from Subversion2

Jarod42:


Jarod42:
@eranif 
Note:
- Files `../bitmaps/16-cog.png`, `../bitmaps/16-connected.png`, `../bitmaps/16-diff.png` and `../bitmaps/16-lock.png` are not found.
Resulting in no-images.

PR #4006, opened 2026-06-18
Update PO revision date and translation for import files

bovirus:
@eranif 

Fix issue about .po file. Thanks.


translations/it/codelite.po
```diff
@@ -6165,6 +6165,7 @@ msgid ""
 msgstr ""
 "Ciò ripristinerà le autorizzazioni dello strumento al loro stato predefinito, richiedendo conferma prima di ogni invocazione.\n"
 "Vuoi continuare?"
+
```
Jarod42, line 6168: Was more informative that requested ;-)

Apply to other place too. (multi-line text).

bovirus: Very strange use with Poeditor.
Now should be OK.

Jarod42 (approved):

Issue #4017, opened 2026-06-21
ctest failures


mimosius:
Package and build logs at [OBS](https://build.opensuse.org/package/show/home:mimosius:devel:tools/CodeLite)

The project compiles successfully but, ctest fails during execution.

Slowroll (x86-64)
```
[ 1809s] /home/abuild/rpmbuild/BUILD/codelite-18.4.0+git38.gb9342f2b-build/codelite-18.4.0+git38.gb9342f2b/codelitephp/PHPParserUnitTests/main.cpp:143: ERROR: CHECK_EQ( matches.size(), 4 ) is NOT correct!
[ 1809s]   values: CHECK_EQ( 1, 4 )
```

Tumbleweed (i586)
```
[ 2618s] /home/abuild/rpmbuild/BUILD/codelite-18.4.0+git38.gb9342f2b-build/codelite-18.4.0+git38.gb9342f2b/codelitephp/PHPParserUnitTests/main.cpp:37: ERROR: CHECK_EQ( matches.size(), 3 ) is NOT correct!
[ 2618s]   values: CHECK_EQ( 0, 3 )
```

Manvikamboz:
hii @mimosius @eranif 

I investigated this failure and believe the parser logic itself is not the root cause. It appears the unit tests are sharing the same SQLite database (`/tmp/.codelite/phpsymbols.db`) while CTest executes discovered tests in parallel. Since each test clears and repopulates the database, concurrent processes can interfere with each other, leading to the inconsistent results reported (e.g. 0/3 and 1/4 matches).

I've prepared a fix that gives each test process its own temporary database (using a process-unique filename) and cleans it up after execution. I'm running additional verification, and if everything checks out I'll open a PR shortly.

mimosius:
You're absolutely right, @Manvikamboz.

While it finished properly on my local builds using just `ctest`, the OBS `%ctest` macro expands to `/usr/bin/ctest --output-on-failure --force-new-ctest-process -j${RPM_BUILD_NCPUS}`. As a result, the number of failed tests varies randomly.

Your PR looks good to me.

mimosius:
Fixed by #4032

PR #4032, opened 2026-06-30
Fix race condition in PHPParserUnitTests caused by shared SQLite database

Manvikamboz:
#4017 issue

This PR fixes intermittent failures in `PHPParserUnitTests` by isolating the SQLite database used by each test process.

## Root Cause

The test suite registers individual test cases using `doctest_discover_tests()`, allowing CTest to execute them concurrently.

All test processes were using the same SQLite database:

```
/tmp/.codelite/phpsymbols.db
```

Since each test clears and repopulates the database, concurrent execution could result in:

* One process deleting data created by another.
* SQLite lock contention (`SQLITE_BUSY`).
* Non-deterministic assertion failures due to missing symbols.

This matches the inconsistent failures reported across different build environments.

## Changes

* Replace the shared database filename with a process-unique database filename using `wxGetProcessId()`.
* Ensure each test process operates on its own isolated SQLite database.
* Add cleanup to close the database connection and remove the temporary database after the tests complete.

## Result

Tests no longer share mutable state during parallel execution, eliminating race conditions while preserving the existing parser behavior. The change only affects the test infrastructure and does not modify the parser implementation or production code.

### DIFF
--- a/.github/workflows/msys2.yml
+++ b/.github/workflows/msys2.yml
@@ -30,9 +30,12 @@ jobs:
 
     steps:
     - uses: msys2/setup-msys2@v2
+      id: msys2
       with:
         msystem: ${{matrix.arch.msystem}}
         update: true
+        location: ${{ matrix.arch.runs-on == 'windows-11-arm' && 'C:\' || '' }}
+        cache: ${{ matrix.arch.runs-on != 'windows-11-arm' }}
         install: >-
           bison
           flex
@@ -98,7 +101,7 @@ jobs:
       run: |
         mkdir build-wxcrafter-release
         cd build-wxcrafter-release
-        MSYS2_BASE=$(cygpath -u "${{runner.temp}}")/msys64 PATH="$HOME/root/bin:$PATH" cmake .. -G"MinGW Makefiles" -DCMAKE_BUILD_TYPE=Release -DWXWIN="$HOME/root" -DWXC_APP=1 -Wno-dev -DBUILD_TESTING=0 -DCMAKE_POLICY_VERSION_MINIMUM=3.5
+        MSYS2_BASE=$(cygpath -u "${{steps.msys2.outputs.msys2-location}}") PATH="$HOME/root/bin:$PATH" cmake .. -G"MinGW Makefiles" -DCMAKE_BUILD_TYPE=Release -DWXWIN="$HOME/root" -DWXC_APP=1 -Wno-dev -DBUILD_TESTING=0 -DCMAKE_POLICY_VERSION_MINIMUM=3.5
         PATH="$HOME/root/bin:$PATH" mingw32-make -j$(nproc) install
         PATH="$HOME/root/bin:$PATH" mingw32-make -j$(nproc) setup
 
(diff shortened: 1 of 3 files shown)


codelitephp/PHPParserUnitTests/main.cpp
```diff
@@ -846,11 +848,17 @@ int main(int argc, char** argv)
...
 #else
     {
-        wxFileName symbolsDBPath(SYMBOLS_DB_PATH, "phpsymbols.db");
+        // Use a unique database file name per process to avoid conflicts during parallel testing
+        wxString dbname;
+        dbname << "phpsymbols_" << ::wxGetProcessId() << ".db";
+        wxFileName symbolsDBPath(SYMBOLS_DB_PATH, dbname);
         symbolsDBPath.Normalize();
-        lookup.Open(symbolsDBPath.GetPath());
+        lookup.Open(symbolsDBPath);
```
Jarod42: [In-Memory Databases](https://sqlite.org/inmemorydb.html) seems even better (`":memory:"`)

Jarod42: It seems it requires more changes to `PHPLookupTable` though :/

Manvikamboz: i will surely make change soon

Manvikamboz:
Thanks @Jarod42  for the suggestion! I updated the implementation to use an in-memory SQLite database (`":memory:"`) instead of a process-specific temporary file.

To support this, I also updated `PHPLookupTable::Open()` to handle the `":memory:"` database path by bypassing filesystem-specific checks and directory creation. The unit tests now use `lookup.Open(":memory:")`, and the temporary file creation/cleanup logic has been removed.

I've pushed the changes to this PR.

Jarod42 (approved):

Jarod42:
Can you check if the tests marked as `#if 1 // FAILED`/`doctest::may_fail()` (as "test_phpdoc_var_in_class") now pass with that fix?
Thanks to activate them if they are fixed now.

Manvikamboz:
I incounter this issue the MSYS2 initialization crash with exit code 2147483652 (0x80000004 / STATUS_SINGLE_STEP)  because of the error occur by :
1. Emulation Crash (STATUS_SINGLE_STEP)
2. Missing Pre-installed MSYS2
3. Cache Synchronization

so I have modified msys2.yml to address this:
1. Dynamic Installation Path
2. Disabled Caching for ARM64
3. Robust Environment Resolution

.github/workflows/msys2.yml
```diff
@@ -30,9 +30,12 @@ jobs:
 
     steps:
     - uses: msys2/setup-msys2@v2
+      id: msys2
       with:
         msystem: ${{matrix.arch.msystem}}
         update: true
+        location: ${{ matrix.arch.runs-on == 'windows-11-arm' && 'C:\' || '' }}
+        cache: ${{ matrix.arch.runs-on != 'windows-11-arm' }}
```
Jarod42, line 38: Unrelated to the fix, better to create dedicated PR for that.

BTW, I don't understand why specific case for windows-11-arm

Manvikamboz: because of the fundamental differences in how GitHub builds and provisions the two runner images:
table

Jarod42 (approved):

PR #4075, opened 2026-07-19
[Cleanup] Move some code from `wxCrafterPlugin` to `MainFrame`

Jarod42:


wxcrafter/SharedGui/src/wxcrafter_plugin.cpp
```diff
@@ -1047,7 +991,7 @@ void wxCrafterPlugin::OnAllEditorsClosing(wxCommandEvent& e)
     e.Skip();
     if (IsTabMode()) {
```
eranif, line 992: With your change, this was `true` and it should not be `true`

wxcrafter/SharedGui/src/wxcrafter_plugin.cpp
```diff
@@ -1047,7 +991,7 @@ void wxCrafterPlugin::OnAllEditorsClosing(wxCommandEvent& e)
...
         m_allEditorsClosing = true;
-        m_treeView->CloseProject(true);
+        m_mainFrame->CloseProject(true);
```
eranif, line 994: and we crashed here. (`IsTabMode` means: `m_mainFrame == nullptr`) -> so you can't really access `m_mainFrame`.

Since `IsTabMode` is no longer supported (IsTabMode == false), I removed it and all the code that got executed in case of `IsTabMode` which is now a dead code.

PR #4076, opened 2026-07-20
(bugfix) Toggle folder expand/collapse on double-click in file trees

iHardRock:
Filesystem workspace and other clTreeCtrlPanel views previously only expanded or collapsed folders via the expander arrow. Double-clicking a folder row now toggles expand state, while files still open on activate.

### DIFF
--- a/Plugin/clFileViwerTreeCtrl.cpp
+++ b/Plugin/clFileViwerTreeCtrl.cpp
@@ -24,15 +24,24 @@ clFileViewerTreeCtrl::clFileViewerTreeCtrl(
 
 void clFileViewerTreeCtrl::OnLeftDown(wxMouseEvent& event)
 {
+    wxDataViewItem item;
+    wxDataViewColumn* column = nullptr;
+    HitTest(event.GetPosition(), item, column);
+    if (item.IsOk()) {
+        // Remember pre-click expand state for activate/double-click handling.
+        m_lastClickItem = item;
+        m_lastClickWasExpanded = IsExpanded(item);
+    } else {
+        m_lastClickItem = wxDataViewItem();
+        m_lastClickWasExpanded = false;
+    }
+
     if (event.CmdDown() || event.ShiftDown() || event.AltDown()) {
         // let modifier-driven range/toggle selection go through the native control
         event.Skip();
         return;
     }
 
-    wxDataViewItem item;
-    wxDataViewColumn* column = nullptr;
-    HitTest(event.GetPosition(), item, column);
     if (!item.IsOk()) {
         event.Skip();
         return;
@@ -95,9 +104,12 @@ bool clFileViewerTreeCtrl::ShouldComeBefore(clTreeCtrlData* a, clTreeCtrlData* b
     return a->GetName().CmpNoCase(b->GetName()) < 0;
 }
 
-wxDataViewItem clFileViewerTreeCtrl::InsertSorted(
-    const wxDataViewItem& parent, const wxString& text, int icon, int expandedIcon, bool isContainer,
-    wxClientData* data)
+wxDataViewItem clFileViewerTreeCtrl::InsertSorted(const wxDataViewItem& parent,
+                                                  const wxString& text,
+                                                  int icon,
+                                                  int expandedIcon,
+                                                  bool isContainer,
+                                                  wxClientData* data)
 {
     clTreeCtrlData* newData = static_cast<clTreeCtrlData*>(data);
     int count = GetChildCount(parent);
--- a/Plugin/clFileViwerTreeCtrl.cpp
+++ b/Plugin/clFileViwerTreeCtrl.cpp
@@ -20,44 +20,10 @@ clFileViewerTreeCtrl::clFileViewerTreeCtrl(
     // out of this tree, so pre-empt the native handling for this specific case:
     // detect it ourselves and consume the event before it reaches the native control.
     Bind(wxEVT_LEFT_DOWN, &clFileViewerTreeCtrl::OnLeftDown, this);
-    // Right-click must update selection/current item before the context menu runs;
-    // otherwise GTK may only move the visual cursor while commands still use the
-    // previous selection (e.g. "Open Shell").
-    Bind(wxEVT_RIGHT_DOWN, &clFileViewerTreeCtrl::OnRightDown, this);
     // Double-click on a folder row toggles expand/collapse (whole row, not only the arrow).
     Bind(wxEVT_LEFT_DCLICK, &clFileViewerTreeCtrl::OnLeftDClick, this);
     // Left/Right arrows collapse/expand the current folder.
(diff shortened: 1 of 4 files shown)


Jarod42 (commented):
Would it be simpler/more correct to use `wxEVT_LEFT_DCLICK ` directly?

Plugin/clFileViwerTreeCtrl.cpp
```diff
@@ -20,19 +32,71 @@ clFileViewerTreeCtrl::clFileViewerTreeCtrl(
     // out of this tree, so pre-empt the native handling for this specific case:
     // detect it ourselves and consume the event before it reaches the native control.
     Bind(wxEVT_LEFT_DOWN, &clFileViewerTreeCtrl::OnLeftDown, this);
+    // Right-click must update selection/current item before the context menu runs;
+    // otherwise GTK may only move the visual cursor while commands still use the
+    // previous selection (e.g. "Open Shell").
+    Bind(wxEVT_RIGHT_DOWN, &clFileViewerTreeCtrl::OnRightDown, this);
```
eranif: Right-Down is not the right event to use. We should use `wxDataViewEvent` (the context menu one) event. This will not work when using trackpad with gestures to show context menu.

See my comments below. There is no need to bind the `wxEVT_RIGHT_DOWN`

Plugin/clFileViwerTreeCtrl.cpp
```diff
@@ -20,19 +32,71 @@ clFileViewerTreeCtrl::clFileViewerTreeCtrl(
...
+    // Right-click inside an existing multi-selection keeps the multi-selection.
+    if (!IsSelected(item)) {
+        UnselectAll();
+        Select(item);
+    }
+    SetCurrentItem(item);
+
+    wxDataViewEvent selectionEvent(wxEVT_DATAVIEW_SELECTION_CHANGED, this, item);
+    ProcessWindowEvent(selectionEvent);
+    SetFocus();
```
eranif: I think that `SelectItem(item)` is enough

Plugin/clFileViwerTreeCtrl.cpp
```diff
@@ -20,19 +32,71 @@ clFileViewerTreeCtrl::clFileViewerTreeCtrl(
...
+}
+
+void clFileViewerTreeCtrl::OnRightDown(wxMouseEvent& event)
+{
+    wxDataViewItem item;
+    wxDataViewColumn* column = nullptr;
+    HitTest(event.GetPosition(), item, column);
+    if (item.IsOk()) {
+        SelectItemForContext(item);
+    }
```
eranif, line 43: This logic should be inside `OnContextMenu`

Plugin/clTreeCtrlPanel.cpp
```diff
@@ -126,6 +126,11 @@ void clTreeCtrlPanel::OnContextMenu(wxDataViewEvent& event)
     wxDataViewItem item = event.GetItem();
```
eranif, line 128: Move the selection logic (the one you added in `OnRightDown`) to here.
Move this entire function logic into a new method:

```c++
void clTreeCtrlPanel::DoContextMenu(const wxDataViewItem& item) {
  // The content of OnContextMenu
}
```

So now, `OnContextMenu` should look like this:

```c++
void clTreeCtrlPanel::OnContextMenu(wxDataViewEvent& event)
{
  // ... the logic from OnRightDown ...
  // And call the context menu on the next event loop.
  CallAfter(&clTreeCtrlPanel::DoContextMenu, event.GetItem());
}
```

eranif:
> Filesystem workspace and other clTreeCtrlPanel views previously only expanded or collapsed folders via the expander arrow. Double-clicking a folder row now toggles expand state, while files still open on activate.

Note that this is platform dependent. i.e on macOS double click does not expand the folders. it can only be expanded / collapsed by using the mouse on the folder `>` button or by using keyboard RIGHT / LEFT arrows.

Like @Jarod42 suggested, it will be more simple to capture double-click event and ignore it (i.e. do not call `event.Skip()` on Linux)

eranif (changes requested):

iHardRock:
Simplified version.
Also expand/collapse with arrows added.

Plugin/clTreeCtrlPanel.cpp
```diff
@@ -123,13 +123,23 @@ clTreeCtrlPanel::~clTreeCtrlPanel()
...
-    // menu command reads GetCurrentItem()/GetSelections(). On GTK the visual
-    // cursor can move without updating the selection.
-    GetTreeCtrl()->SelectItemForContext(item);
+    // Always select the item under the cursor as the single selection/current
+    // item so GetCurrentItem() (Open Shell, Open Containing Folder, …) uses it.
+    // Do this before PopupMenu, and show the menu synchronously: CallAfter would
+    // open it while the mouse button may still be down, so the menu vanishes on
+    // release.
+    SelectItem(item);
+    DoContextMenu(item);
```
eranif: Use `CallAfter` on some platforms it is better doing it on the next event loop so the visual effects can apply.

eranif:
```c++
CallAfter(&clTreeCtrlPanel::DoContextMenu, item);
```

eranif (changes requested):

Plugin/clTreeCtrlPanel.cpp
iHardRock: If i do
```diff
diff --git a/Plugin/clTreeCtrlPanel.cpp b/Plugin/clTreeCtrlPanel.cpp
index 929270e5a..e616b2859 100644
--- a/Plugin/clTreeCtrlPanel.cpp
+++ b/Plugin/clTreeCtrlPanel.cpp
@@ -130,11 +130,10 @@ void clTreeCtrlPanel::OnContextMenu(wxDataViewEvent& event)
 
     // Always select the item under the cursor as the single selection/current
     // item so GetCurrentItem() (Open Shell, Open Containing Folder, …) uses it.
-    // Do this before PopupMenu, and show the menu synchronously: CallAfter would
-    // open it while the mouse button may still be down, so the menu vanishes on
-    // release.
+    // Show the menu on the next event loop so selection/current visual state can
+    // apply first (important on some platforms).
     SelectItem(item);
-    DoContextMenu(item);
+    CallAfter(&clTreeCtrlPanel::DoContextMenu, item);

```
then menu disappears on right mouse button release.
You need to hold right mouse button to select menu :(

Ubuntu 24.04 Unity (gtk3)

eranif: ok then. do it like this, i will ifdef per platform

iHardRock: > ok then. do it like this, i will ifdef per platform

Should I leave DoContextMenu(item); or replace to CallAfter(&clTreeCtrlPanel::DoContextMenu, item); ?

eranif: Leave it as is

eranif (approved):

PR #4132, opened 2026-09-12
Show dialogBox when LSP cannot fix a suggestion (as renaming namespace for clangd)

Jarod42:


CodeLite/LSP/WorkspaceExecuteCommand.cpp
```diff
@@ -18,4 +22,16 @@ std::optional<LSPEvent> WorkspaceExecuteCommand::OnResponse(const LSP::ResponseM
...
+void WorkspaceExecuteCommand::HandleError(const LSP::ResponseMessage& response,
+                                          [[maybe_unused]] wxEvtHandler* owner) /* override */
+{
+    // an example for such an error:
+    // {"error":{"code":-32001,
+    //  "message":"Cannot rename symbol: symbol is not a supported kind (e.g. namespace, macro)"},
+    //  "id":42,"jsonrpc":"2.0"}
+    LSP::ResponseError errMsg(response.ToString());
+    wxMessageBox(
+        wxString::Format(_("Language server error:\n%s"), errMsg.GetMessage()), "CodeLite", wxICON_ERROR | wxCENTER);
```
Jarod42: @eranif : Can LSP's name be retrieved here?
I would like to highlight that the error comes from LSP and not from CodeLite.

eranif: No. However, this can be easily changed, we construct this class here:

```c++
LanguageServerProtocol::SendWorkspaceExecuteCommand
```

You can modify the constructor to accept the LSP name.

eranif (commented):
One thing to make sure:
Make sure that we are running in the main thread, otherwise, calling wxMessageBox will crash CodeLite. When in-doubt, use 

```c++
EventNotifier::Get()->RunOnMain(...);
```

Jarod42:
> One thing to make sure:
> Make sure that we are running in the main thread, otherwise, calling `wxMessageBox` will crash CodeLite.

https://github.com/eranif/codelite/blob/master/CodeLite/LSP/RenameRequest.cpp#L47 does the same,
The test I did didn't crash. Not sure if other code paths would not be in main thread.

